PR #2, opened 2020-03-21
Redesign

alessandrofavero:
Redesign of the app following @Mattro design.
As a starting point for further development.


maks (approved):
Great work!

PR #3, opened 2020-03-22
Home page tuning

alessandrofavero:
Tuned the home page to match @Mattro  design. 
Added icon for both iOS and Android following @Mattro files.
Commented out testing, @maks saying it was generating errors on codemagic build.


test/widget_test.dart
```diff
@@ -5,26 +5,10 @@
 // gestures. You can also use WidgetTester to find child widgets in the widget
 // tree, read text, and verify that the values of widget properties are correct.
 
-import 'package:flutter/material.dart';
-import 'package:flutter_test/flutter_test.dart';
+// import 'package:flutter/material.dart';
```
lukesleeman, line 8: I'm definately no expert on flutter, but it might be a good idea just to delete these rayther than leave them in commented out.  We can always get the history back out of git if we need :-)

lukesleeman:
Hey @alessandrofavero Its really exciting seeing this come in.  Is it possible for you to paste some designs in this PR so we can see what the app looking like?

You might also want to add @adamkoch, @workerbee22 and brett (once he joins the project) to the list of reviewers.  You only need one ✅ to be able to merge

test/widget_test.dart
workerbee22: Actually need flutter_test very soon to pass CI testing.

workerbee22:
Its ok for me .... I'm looking at the code in AS and running locally, so I can see exactly what the app looks like in Android and iOS 👍

test/widget_test.dart
alessandrofavero: I commented it out because it was causing some build problems and I wasn't using it.  But I knew that someone might need it.

workerbee22: That will be because there are no tests yet, but that sall good. Leave it for now 👍I can fix later and add at least one test so it gets past build on codemagic.

maks: Yes lets please not leave any code, even tests commented out. We'll start adding tests asap

lib/widget/card_container.dart
```diff
@@ -44,18 +45,40 @@ class CardContainer extends StatelessWidget {
     switch (cards.length) {
       case 2:
         return Row(
-          children: expandedCards,
+          children: <Widget>[
+            SizedBox(
+              width: 16,
```
maks, line 50: lets try to get away from using hard coded sizes as soon as we can, but its ok for now

maks (approved):
:+1:  Good to merge - just please keep an eye on build passing before merging as for some reason CI build results not showing up here on the PR yet (I'll get that fixed asap)

lib/widget/card_container.dart
alessandrofavero: I know @maks, but I was having some problems rendering properly. So, for now, let's use SizedBox a little bit.

PR #5, opened 2020-03-22
Added Inter Font from @mattro

alessandrofavero:
Added Inter font to the app as a global font.


pubspec.lock
```diff
@@ -101,7 +101,7 @@ packages:
       name: pedantic
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "1.9.0"
+    version: "1.8.0+1"
```
maks: @alessandrofavero not sure why this version is being downgraded? 1.9 is latest in pub.dev - are you on latest flutter stable channel?

maks (changes requested):
just one small nit about pubspec downgrading pendatic package

pubspec.lock
alessandrofavero: Yes I am. That's weird I didn't touch pubspec.yaml other than adding the font. I'll fix it.

alessandrofavero: All fixed in the latest commit.

maks (approved):
:+1:

PR #10, opened 2020-03-22
Add simple info view

adamkoch:
Add the simple info view screen (when you click the little "info" button on the home screen)

## Changes

Adds new screen. Modifies reusable_card. See below for other notes from changes.

## Screenshots

Screen Shot 2020-03-22 at 11 46 48 pm

## Things to note
 - adds dep on package_info package
 - bumps Dart version to min 2.3.0 (to allow for some newer features like spread)
 - adjusts reusable_card to support a single row type mode (when description is null), it also removes the sizedbox in the middle in favor of setting the height of the overall card (found some inconsistent height with cards that had title wrap over otherwise)

### DIFF
--- a/lib/hard_data.dart
+++ b/lib/hard_data.dart
@@ -52,6 +52,7 @@ final List<ReusableCard> icu = [
   ),
   ReusableCard(
     title: 'Tips for Junior Staffers',
+    description: '',
     color: Color.fromRGBO(143, 217, 255, 1),
   ),
 ];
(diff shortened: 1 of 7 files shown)


alessandrofavero:
Hi Adam, at a first glance (I'm at work now sorry) looks all good to me! I've added @maks and @workerbee22 as reviewers so they can approve it.

maks (approved):
Wow that was quick, thanks :+1: looks great Adam!

PR #11, opened 2020-03-22
Tidying up minor nits

domesticmouse:
Tidying up minor concerns in the code base

## Changes

- Make analysis_options unversioned (always use latest)
- Bump Dart min version to 2.7 as this codebase uses list if and splice
- Move out of com.example and into org.whegba
- Add web target for testing purposes
- Add whitespace to bottom of list to give bottom card room to breathe

## Screenshots

Before:
WHEGBA Before

After:
WHEGBA After


maks (approved):
:tada: :+1:

PR #13, opened 2020-03-22
Info view tweaks, clean up routing consts

adamkoch:
Bit of clean up for info view.

## Changes

- Change background color
- Switch to full screen dialog
- Move app routes into own class


maks (approved):
:+1:  :100: on putting Routes into own consts

PR #52, opened 2020-03-23
Include new assets from Marc

lukesleeman:
This PR includes new assets from @marcedwards 

Huge caveat:

image

I am not a flutter dev, I'm just chucking them in the codebase so we have them.  If they are in the wrong place can somebody who actually knows what they are doing move them around on this branch and get this PR over the line?

## Screenshots

image

## Things to note

Marc says:
> A note about those assets: They’re red to ensure the tinting is working correctly. I’m assuming tinting is possible and we’re likely to use it?


adamkoch:
PR looks good. I think red color should work for blend mode but I haven't tested before (usually just white icon on transparent background but I think any opaque color can be blended).


Also, what is the source for these icons? Custom? There's a few flutter packages that have common icon packs/fonts available in case the icons are from one of those, eg:
https://github.com/flutter/cupertino_icons
https://github.com/fluttercommunity/font_awesome_flutter (https://fontawesome.com/)

cc/ @marcedwards

adamkoch (changes requested):
I think 1.0x assets can just go straight in the icon/ dir (1.0x is the default).

alessandrofavero:
What about creating a custom icon font with https://fluttericon.com/ ? So we can use it as Icon() and access all the properties like color, size .... @marcedwards

ericjiang97:
Like the icon font idea!

suesitran:
I can do the conversion during lunch time :D but the icon color is always black. Let me know if this is okie. This means we'll need to keep updating color when we use these icons.

**Color code:** 
HTML/HEX code: | #fc6665 |  
-- | -- | --
RGB code: | rgb(252, 102, 101) |  
Hsl code: | hsl(0, 96%, 69%) |  
Hwb code: | hwb(0, 40%, 1%) |  
Cmyk code: | cmyk(0%, 60%, 60%, 1%) |  
Ncol code: | R1, 40%, 1% |

marcedwards:
I don’t think we should use an icon font. There’s not much to be gained from it, and we already have the icon sizes we need. As it stands, we can already tint the PNG images. An icon font adds additional authoring steps, which will slow us down.

suesitran:
> An icon font adds additional authoring steps, which will slow us down.

The problem with adding Icon fonts I can see is that:
1. need to update font to include new and old icon every time we add new icon, or generate new font files, which will create more font files as we grow
2. time to generate icon font files
3. all icons are black, we need to tint the color whenever we use IconData
4. adjust size if we need bigger or small icon

I like the idea of icon fonts, but it's not feasible atm.

lukesleeman:
As previously mentioned, I have no idea what I'm doing 😁

I've assigned this PR to @suesitran  - she can make whatever changes are necessary to get this in and merged.

suesitran:
Remove PNGs, keep only SVGs for future conversions if we need to add more icons as IconData

maks (changes requested):
@suesitran Looks almost ready. Could you just remove the change in the pubspec.lock file and the ttf file as I agree with Marc there really no big advantage at this point for have icon font.

lukesleeman:
Thanks @suesitran for all your hard work on this ❤️ 

Are we able to re-review and merge now?

maks (approved):
:+1: :tada: Thanks @suesitran !

suesitran:
@adamkoch anything else u need me to change?

adamkoch (approved):

adamkoch:
LGTM, thanks @suesitran!

Issue #53, opened 2020-03-23
Embedded video to change in Putting on Personal Protective Equipment


greggmiller:
Hi Adam,

Minor (hopefully) amendment required. 
The video/link currently embedded in putting on PPE is the one for taking it off.

Could you switch to video from this link please: 
https://www.youtube.com/watch?v=jZyR0S8FKIY&feature=youtu.be

Let me know if any queries. Cheers!

adamkoch:
Yep will do.

adamkoch:
Fixed in #55

Issue #30, opened 2020-03-23
Build Your Welfare Screen


lukesleeman:
Under "Look after yourself" the following subscreen needs building:

<img width="289" alt="image" src="https://user-images.githubusercontent.com/902252/77297264-20dcdf00-6d3d-11ea-9bf4-94cf5335cced.png">

for the "Phases of Pandemic and suggestions for self-care" button, please link it out to a subscreen which can contain the text "TODO"

lukesleeman:
👋 @greggmiller do you know what phone number the phone icon on this screen should go to?

adamkoch:
Initial build done. Still need correct phone number and to put in the icons.

lukesleeman:
@adamkoch Is the phone number you need for the little Phone icon?

I would say don't worry about that now - the phone buttons on each screen will a feature that comes along later.  I'll fire up a new issue for doing that a little later this evening.

Wondering if we could close out this issue now?

adamkoch:
Ah got it, actually @alessandrofavero updated the phone icon to go to the contacts web link already in https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/pull/58 so we're all set there.

Can close this one out once this is merged:
https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/pull/61

Issue #21, opened 2020-03-23
Add typography styles to style.dart


marcedwards:
This should be all or most of the text styles we need in the app. They’re all named similarly to HTML elements (H1, H2, P etc). I’m prefixing everything with text. Please let me know if that’s a good idea or not.

Please note that these styles reference the colour constants we have (listed in #9).

![textstyles](https://user-images.githubusercontent.com/1121547/77273222-21f31980-6d07-11ea-8c86-d3030b774e02.png)

```
const TextStyle textH1 = TextStyle(
  color: blackAlpha900,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.bold,
  fontSize: 28.0,
  fontFamily: 'InterBold'
);

const TextStyle textH2 = TextStyle(
  color: blackAlpha900,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.regular,
  fontSize: 28.0,
  fontFamily: 'Inter'
);

const TextStyle textH3 = TextStyle(
  color: blackAlpha900,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.bold,
  fontSize: 22.0,
  fontFamily: 'InterBold'
);

const TextStyle textH4 = TextStyle(
  color: blackAlpha900,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.regular,
  fontSize: 22.0,
  fontFamily: 'Inter'
);

const TextStyle textH5 = TextStyle(
  color: blackAlpha900,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.semibold,
  fontSize: 17.0,
  fontFamily: 'InterSemiBold'
);

const TextStyle textP = TextStyle(
  color: blackAlpha600,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.regular,
  fontSize: 17.0,
  fontFamily: 'Inter'
);

const TextStyle textFooter = TextStyle(
  color: blackAlpha600,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.regular,
  fontSize: 13.0,
  fontFamily: 'Inter'
);

const TextStyle textCaption = TextStyle(
  color: blackAlpha600,
  fontWeight: FontWeight.w600,
  fontStyle: FontStyle.semibold,
  fontSize: 13.0,
  fontFamily: 'InterSemiBold'
);
```

### Questions and notes

[...]

PR #55, opened 2020-03-23
Put in text styles, fix doffing video, add welfare screen

adamkoch:
Adds new typography styles #21 
Adds initial welfare screen #30 (although still needs proper icons)
Fixes doffing video file #53 

## Screenshots

Screen Shot 2020-03-24 at 7 56 10 am


ericjiang97 (approved):
Lgtm! 🔥 🚀

pubspec.yaml
```diff
@@ -34,5 +35,17 @@ flutter:
   fonts:
     - family: Inter
       fonts:
-        - asset: assets/fonts/Inter.ttf
+        - asset: assets/fonts/Inter-300-Light.ttf
```
domesticmouse, line 38: nit: consider using package:google_fonts instead of hard including inter.

More detail: https://pub.dev/packages/google_fonts

domesticmouse (approved):

pubspec.yaml
adamkoch: Downside of google fonts is it requires http download one time before fonts load. Not a fan of that.

domesticmouse: Fair enough. Have you raised a bug on `google_fonts` about the http download requirement? I'll happily advocate for a build time option with the eng team.

adamkoch: I haven't raised it. I assumed it was WAI as it gives you easy access to the online library of Google fonts (major plus) with the downside it needs to do a small download upon startup. Using local assets seems to be a fine alternative if you just want a small number of fonts with no http download (downside being increased binary size).

domesticmouse: Fair. I'll still bug the team tho =)

PR #75, opened 2020-03-25
Fix hand hygiene PPE images

marcedwards:
This adds a duplicate hand hygiene image, so we have one for donning PPE (putting on), and one for doffing PPE (taking off). Having duplicate images seems safer than referencing a donning image in the doffing instructions.


lukesleeman (approved):
Good call on this, it Looks good to me

✅ 🚢

lukesleeman:
Given Kim needs this for her task and its approved, I'm going to click Merge 😁

PR #84, opened 2020-03-25
Adding code style lint checks

domesticmouse:
Tidying up some code issues via the joys of codelints.

## Changes

- Flutter format
- Add `const`s in places to make things faster
- Use list if instead of ternary operator 

## Things to note

Use of `const` makes flutter run faster, because compile time constant data always pointer compares equal to itself, which makes flutter's diff algorithm take a lot less time for constant subtrees.


lib/main.dart
```diff
@@ -1,20 +1,19 @@
...
-import 'package:wh_covid19/view/info_view.dart';
-import 'package:wh_covid19/view/intubation/algorithm/intubation_algorithm_page.dart';
-import 'package:wh_covid19/view/intubation/checklist/intubation_checklist_page.dart';
-import 'package:wh_covid19/view/intubation/guidance/intubation_guidance_page.dart';
-import 'package:wh_covid19/view/ppe/ppe_view.dart';
-import 'package:wh_covid19/view/sbs_guide_view.dart';
-import 'package:wh_covid19/view/view_templates/html_text_card_view_template.dart';
-import 'package:wh_covid19/view/staff_welfare/your_welfare_view.dart';
-
+import 'hard_data.dart';
```
maks, line 2: @domesticmouse I thought the recommendation was to use package imports rather than relative imports?

maks (commented):
@domesticmouse Great work with the consts! Just have one comment about the change to relative imports...

lib/main.dart
domesticmouse: Uh, where did you get that from? (It actually leads to the type system getting confused, because it thinks the package local type and the external package import are two different types.)

I'm genuinely curious where you picked up that rule of thumb from, because I need to go correct it, either by editing our docs, or sending someone an email asking them to change their docs.

adamkoch: Interesting! Not sure I read it anywhere, but I personally have preferred it for consistency.

Also:
https://github.com/flutter/flutter/blob/86389be66e8e0bd3e082b20a6bb3251d7c7706d5/analysis_options.yaml#L73
🤷

Although it does say to prefer relative here:
https://dart.dev/guides/language/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory
But then it says: "There is no profound reason to prefer the former—it’s just shorter, and we want to be consistent."

But then it does say:
An import path should never contain /lib/
A library under lib should never use ../ to escape the lib directory.

So I guess relative path like "view/home_page.dart" is recommended, but a relative path like "import '../hard_data.dart';" is not recommended. At least according to the docs 🤔

domesticmouse: It's something I'm guessing Java/Android developers will do because it feels right. 

For better or worse, in Dart path relative imports and package imports mean different things to the Dart's type system. Thus the lint checks =)

adamkoch: > It's something I'm guessing Java/Android developers will do because it feels right.

That explains why I have a leaning toward it :)

This PR still has '../' style imports so I think those should be removed as per the dart.dev guide?

domesticmouse: The `../` style imports are not escaping the `lib` directory. They are just traversing within the `lib` directory, and thus are totes kosher.

adamkoch: Ah ok got it. Thanks for the clarification.

adamkoch: Also, if you're saying there is a "real" reason to use relative versus package, you may want to ask them to update the docs which say:
"There is no profound reason to prefer the former—it’s just shorter, and we want to be consistent."

adamkoch: Also:
https://github.com/dart-lang/pedantic/blob/master/lib/analysis_options.1.9.0.yaml#L16

adamkoch (approved):

lib/main.dart
domesticmouse: Yup, I've got that lint included as well: 

https://github.com/domesticmouse/wh_covid19_app/blob/Adding-code-style-checks/analysis_options.yaml#L28

domesticmouse: Actually, I should be getting a warning that it is double included. Meh.

maks:
@domesticmouse with relative vs package imports I remembered reading this issue: https://github.com/flutter/flutter-intellij/issues/2015
But maybe I got @zoechi's advice backwards and you should always use relative within a package and so within a flutter app too?

ericjiang97 (approved):

domesticmouse:
@maks this PR never gets a clean bill of health from codemagic, I'd love some insight as to what is going wrong. Thanks!

maks:
sure thing @domesticmouse I'll have a look into it now!

domesticmouse:
Thanks @Maks 😁

domesticmouse:
Replaced by #95

lib/main.dart
maks: @domesticmouse it seems to be used all over the place in Googles own samples:

https://github.com/flutter/samples/tree/master/veggieseasons/lib/main.dart
https://github.com/flutter/gallery/blob/master/lib/main.dart
https://github.com/flutter/samples/blob/master/provider_shopper/lib/main.dart
etc

but I think this maybe due a old bug since fixed as most recent comments on Remi's answer: https://stackoverflow.com/questions/47121411/flutter-retrieving-top-level-state-from-child-returns-null/47142052#47142052

zoechi:
@maks that comment was related to a quite annoying Dart bug that was fixed about a year ago. The bug only occured when relative imports were used in the `lib/main.dart` file. This file is specific to Flutter and therefore this bug couldn't happen in plain Dart projects. Otherwise it would probably have been fixed much earlier.

domesticmouse:
Thanks for the clarification @zoechi =)

PR #87, opened 2020-03-25
Steps to update release notes in PR template

lukesleeman:
This adds a checkbox to the PR template for updating the release-notes.  I've also filled out the notes based on going back through our git history

## Changes

- Added checkbox to pr template
- Filled out releas notes

## Things to note

Not 100% sure if this is the structure we want for the release notes?  Given we are shipping a build with every merge to master, I figured its best to just group changes by day.  People can add a new day to the list as needed.

### DIFF
--- a/release-notes.txt
+++ b/release-notes.txt
@@ -1,1 +1,30 @@
-Initial app version.
+---- Thursday 27th March, 2020 ----
+
+
+---- Wednesday 25th March, 2020 ----
+
+* Contact numbers now shows WH phonebook webpage
+* Added initial versions of ICU screens
+
+---- Tuesday 24th March, 2020 ----
+
+* Improved PPE screens including videos
+* Built out your welfare screen
+* Build out rough versions of intubation screens, including intubation steps infographic with pinch to zoom
+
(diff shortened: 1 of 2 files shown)


release-notes.txt
```diff
@@ -1 +1,30 @@
-Initial app version.
\ No newline at end of file
+---- Thursday 27th March, 2020 ----
```
lukesleeman, line 1: @maks and @workerbee22 I'm not entirely sure about this format?  I'm hoping it will work out well enough in firebase app dist?

Feel free to suggest changes to this as you see fit

maks: @lukesleeman  I'd add the need to update the release notes on both App Store and Google Play listings as I dont know if there is any automation available in Codemagic to upload that as part of process of distributing to stores, though @workerbee22 may know of a script to do that.

maks (approved):
:+1: @lukesleeman don't forget to include [skip ci] in your doco-only commits to save having to use build minutes on doc only changes :slightly_smiling_face: We can always override the missing build status on these kinds of PRs.

PR #104, opened 2020-03-27
add initial tests for init router, disclaimer

maks:
Add tests for on start screen (init router) and showing disclaimer and showing homescreen.


ericjiang97 (approved):
Looks good :)

Issue #96, opened 2020-03-26
Changes to wording on home screen and on app hierarchy map


greggmiller:
After several discussions throughout today with @Western-Health-Covid19-Collaboration/medical  and others.

**_Changes are:_**

**ICU** 
**Ventilation** remains
**General care** change to **Daily Round**
Tips for **Junior staff**  change to **Cross-skilling**

(Cross-skilling is the likely event that many medics will be working in roles outwith their usual responsibilities during the surge of the pandemic)

 **Airway** 

Change **Algorithm** to **Infographics**

**Info and resources** 
Add **WH Novel Coronavirus information** (previously on hierarchy as Google Covid News Feed) (this is only a link to https://coronavirus.wh.org.au/# so will be fit for devs when in)

On info and resources add **Acknowledgements and references** section

@Western-Health-Covid19-Collaboration/designers 
@Western-Health-Covid19-Collaboration/medical 

I have put as number 1 priority as I think changing this is a quick fix that may be ready for devs depending on @Mattro and  @marc thoughts? It also impacts quite significantly on the understanding and navigation of the app.

I think this aligns with the discussions from today but please feel free to comment/query.

lukesleeman:
Yep, I think either @Mattro or @marcedwards Will need to update the `Main` and `Flow Diagram` screens in zeplin, with these changes, then we should be good to go.

Mattro:
I’ll knock this one off in the next hour.

lukesleeman:
Amazing!  🙌

Post some pics and move to ‘read for dev’ when you get it done 🙂

Luke

Mattro:
This is ready to go @lukesleeman

greggmiller:
Awesome!

@Western-Health-Covid19-Collaboration/flutter-devs 
Over to you! Thanks!

greggmiller:
@Mattro  Sorry after discussing this I’ve realised it might be better if Intubation and Extubation were laid out symmetrically. They are essentially equal and opposite actions (tube goes in, tube comes out!). So it will be more intuitive and functional for the user if the architecture is  the same:

Airway                        -->  Intubation --> Steps, checklist, infographics
                                   --> Extubation --> Steps, checklist, infographics

greggmiller:
Can I retract previous comment, (unless @Western-Health-Covid19-Collaboration/designers @Mattro  have been working on it in last few hours).

It is good for the first iteration if extubation is one box. 

[...]

PR #117, opened 2020-03-29
[Enhancement] Changes the Wording on Home Screen

yimingc2:
Resolved issue #96

## Abstract
This PR will change the wordings on the home screen and add two sections in Info and Resources screen.

## Screenshots
changes

## Changes
- Change **General care** to **Daily Round**
- Change **Junior staff** to **Cross-skilling**
- Change **Algorithm** to **Infographics**
- Add **WH Novel Coronavirus information** section which links to https://coronavirus.wh.org.au/#
- Add **Acknowledgements and references** section

(👆 delete this line if your changes have no user facing features)

## Things to note
At the moment, the user has to scroll down to see more sections on Info and Resources screen. We will improve it later if there's more time.


lib/coronavirusInfo_launcher.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:url_launcher/url_launcher.dart';
+
+class CoronavirusInfoLauncher {
```
rockgecko-development: Should use `contact_launcher_` which already has these functions

lib/hard_data.dart
```diff
@@ -38,7 +38,7 @@ final List<ReusableCard> intubation = [
     routeTo: Routes.intubationChecklist,
   ),
   const ReusableCard(
-    title: 'Algorithm',
+    title: 'Infographics',
     description: '12 steps',
     color: AppColors.backgroundGreen,
     routeTo: Routes.intubationAlgorithm,
```
rockgecko-development, line 44: Should the route also be renamed? And `algorithm\intubation_algorithm_page.dart` -> `infographics\infographics_page.dart`?

rockgecko-development (commented):
Left a few notes on renaming Algorithm->Infographics, and the url launcher class. Otherwise looks good

lib/view/icu_non_intensivist/general_care_view.dart
```diff
@@ -4,7 +4,7 @@ import 'package:flutter/widgets.dart';
 import '../../style.dart';
 
 class GeneralCareView extends StatelessWidget {
-  final title = 'General Care';
```
suesitran, line 7: duplicated change, there's a same string in Hard Data. Are we consider using a string resource file so we can avoid these duplicated strings?

lib/coronavirusInfo_launcher.dart
yimingc2: Yeah, `contact_launcher` also has its own methods such as `makeCall`. So maybe can create a `base_launcher` for `contact_launcher` and `coronavirus_info_launcher`.

lib/hard_data.dart
yimingc2: You are right. I was thinking if I should do it because the route and variable names should be more general otherwise we'd have to change it all the time when the design is changed.

lib/view/icu_non_intensivist/general_care_view.dart
yimingc2: Totally agree. We should have a file to store all the static strings.

yimingc2: Totally agree. We should have a file to store all the static strings.

lukesleeman: Yep, @workerbee22 is on the case!  We should be using a string resource file

yimingc2: If @workerbee22 is on it, then I'll update this part later.

lukesleeman:
I've added some small tweaks to the design over in https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/96#issuecomment-605612891

@yimingc2 are you able to update this PR to include them?

lib/view/info_view.dart
```diff
@@ -18,6 +19,13 @@ class InfoView extends StatelessWidget {
   }
 
   final topCards = <Widget>[
+    ReusableCard(
+      title: 'WH Novel Coronavirus information',
+      color: _cardColor,
+      height: _cardHeight,
+      elevation: _cardElevation,
+      fallback: () => BaseUrlLauncher.launchWithBrowser(whCoronavirusInfoURL),
```
rockgecko-development: I think you might have misunderstood, and have made this unnecessarily complicated. All your code does, is call `if (await canLaunch(url)) await launch(url)`, which already existed in `contact_launcher.dart`. No need to turn one class into three, just use existing `contact_launcher.dart`.

rockgecko-development (changes requested):

lib/hard_data.dart
rockgecko-development: Good point. I guess if it was originally a typo, it should be changed. If it's a design change, then leaving it is ok.

lib/view/info_view.dart
yimingc2: It is bit confusing to use `ContactLauncher` to launch coronavirus information. And `contact_launcher.dart` has functions like `makeCall` and `sendEmail` which are not related to launching coronavirus information. That's why I created a `BaseUrlLauncher` which has the base functions to launch a URL.

rockgecko-development:
Almost, but I'll write the changes I still would like, below:

1. Rename /utils/url_launcher.dart to /utils/url_utils.dart, and have `class UrlUtils { ...` inside. The class name should match the file name (but without the _), and you can't call the class UrlLauncher because that could be confused with the content of  `package:url_launcher`.
2. Delete contact_launcher.dart. Update all references to point to `UrlUtils.launchWithBrowser` etc

Much simpler and easier for everyone to understand. It's a utility class that can be used by any screen, to launch a url, or tel link etc.

yimingc2:
> Almost, but I'll write the changes I still would like, below:
> 
> 1. Rename /utils/url_launcher.dart to /utils/url_utils.dart, and have `class UrlUtils { ...` inside. The class name should match the file name (but without the _), and you can't call the class UrlLauncher because that could be confused with the content of  `package:url_launcher`.
> 2. Delete contact_launcher.dart. Update all references to point to `UrlUtils.launchWithBrowser` etc
> 
> Much simpler and easier for everyone to understand. It's a utility class that can be used by any screen, to launch a url, or tel link etc.

yimingc2:
> Almost, but I'll write the changes I still would like, below:
> 
> 1. Rename /utils/url_launcher.dart to /utils/url_utils.dart, and have `class UrlUtils { ...` inside. The class name should match the file name (but without the _), and you can't call the class UrlLauncher because that could be confused with the content of  `package:url_launcher`.
> 2. Delete contact_launcher.dart. Update all references to point to `UrlUtils.launchWithBrowser` etc
> 
> Much simpler and easier for everyone to understand. It's a utility class that can be used by any screen, to launch a url, or tel link etc.

Ah cool. Just pushed. I thought `makeCall` and `sendEmail` are something specific to contact and not related to urls. Thanks for clarifying.

rockgecko-development (approved):
Looks great!

PR #122, opened 2020-03-29
Bugfix/android disclaimer exit

joshcase:
Small disclaimer behaviour changes

## Changes

- Hitting the back button on Android now closes the app if the disclaimer is not accepted. If the disclaimer has been accepted, it pops the route instead.

## Things to note

- Cannot test this effectively as:
a) I'm not aware of a way to simulate the Android back button in a widget test
b) Calling "exit(0);" makes the test suite behave unusually

Happy to provide testing if anyone can provide insight into a) and b)


rockgecko-development:
Hi Josh,

UPDATE: ignore this, looks like this is a preexisting issue, I've created a new PR for it, sorry! LGTM

-----


After accepting the disclaimer, whenever I hit Android back key from the home_page, I get a weird animation shadow flash, and I'm still on the home page. I have to hit back again to exit.
IE, I have to hit back twice every time to leave the app.

I put some content in the build method of IntroRouter, and I can see that this widget is showing after hitting back the first time.

joshcase:
No worries, sounds like that might be an issue related to having more than one HomePage in the navigator queue somehow.

rockgecko-development (approved):

PR #124, opened 2020-03-29
Fix for needing to hit back twice to exit app

rockgecko-development:
This fixes the need to hit the android back button twice to exit the app.

Issue was the route `/` is special, so I have made `introRouter = '/'`, and `home = '/home'`

Also changed IntroRouter to host either HomePage or DisclaimerView directly, without a nav event.

I have tested this thoroughly on Android, by wiping app data, hitting back from disclaimer screen and from home screen, both the first time app is run, and subsequently after disclaimer is accepted.
Have also tested nav to disclaimer from info screen.

Have not tested if the ios swipe from top left interaction works properly on the disclaimer screeen (ie, does nothing or closes the app when disclaimer is not accepted).

cc @joshcase in #122


lukesleeman:
👋 Just a heads up @workerbee22 is working on fixing things up with the disclaimer screen right now.  You might want to hold off on this until he gets his changes completed

workerbee22:
About to look at this one and impact on the similar one I'm working on. So should not be much longer 😁

workerbee22 (approved):
I'm working on disclaimer a bit more including this fix. I also made very similar changes. Note mine will change routes a little more again, but these are great and are the same changes I did, so I know they work correctly.

workerbee22:
@alessandrofavero IF you can please review this PR as well as soon as you can. My Disclaimer work (which is just commits right now, now PR raised yet) are in effect a superset of these changes to the routes.

lib/intro_router.dart
```diff
@@ -1,32 +1,27 @@
...
+    return FutureBuilder<SharedPreferences>(
+      future: SharedPreferences.getInstance(),
+      builder: (context, snapshot) {
+        bool disclaimerFirstView;
+        if (snapshot.data != null) {
+          disclaimerFirstView =
+              snapshot.data.getBool("disclaimer_first_view") ?? false;
+        }
+        if (disclaimerFirstView == null) {
+          return Container();
```
alessandrofavero, line 18: Instead of a container I would show a sort of loading indicator just in case the future takes time to resolve

lib/intro_router.dart
```diff
@@ -1,32 +1,27 @@
...
-    }
+    return FutureBuilder<SharedPreferences>(
+      future: SharedPreferences.getInstance(),
+      builder: (context, snapshot) {
+        bool disclaimerFirstView;
+        if (snapshot.data != null) {
+          disclaimerFirstView =
+              snapshot.data.getBool("disclaimer_first_view") ?? false;
+        }
+        if (disclaimerFirstView == null) {
```
alessandrofavero, line 17: I would put this check as an else of snapshot.hasData and also I would manage the case of an error with a check snapshot.hasError

lib/intro_router.dart
```diff
@@ -1,32 +1,27 @@
...
-    if (disclaimerFirstView) {
-      await Navigator.of(context).pushReplacementNamed(Routes.home);
-    } else {
-      await Navigator.of(context).pushReplacementNamed(Routes.disclaimer);
-    }
+    return FutureBuilder<SharedPreferences>(
+      future: SharedPreferences.getInstance(),
+      builder: (context, snapshot) {
+        bool disclaimerFirstView;
+        if (snapshot.data != null) {
```
alessandrofavero, line 13: I would compact this with  snapshot.hasData

alessandrofavero (commented):
There are some changes that I would implement to better manage the snapshot, other than that LGTM good job!

workerbee22:
Suggest we push this through as I have other changes for the routes just like this but also more ready to go. It does this fix AND preps for the Onboarding screens that are now Priority #7 

So LGTM. Everyone ready ?

workerbee22:
Kicked off the PR workflow. See how it goes.

workerbee22:
@bramt Pedantic lint checks strike again. CI build errors on this one:

image

lib/intro_router.dart
```diff
@@ -1,32 +1,27 @@
...
-  _IntroRouterState createState() => _IntroRouterState();
-}
-
-class _IntroRouterState extends State<IntroRouter> {
-  @override
-  void initState() {
-    super.initState();
-    _checkDisclaimerFirstView();
-  }
+import 'package:wh_covid19/view/disclaimer_view.dart';
```
lukesleeman: I'm not a flutter dev, but this should fix up the flutter analyze warning

```suggestion
import 'view/disclaimer_view.dart';
```

workerbee22:
@lukesleeman This won't get past analysis until its fixed.  Not sure if @bramt want to do it, or I can jump in later and do it.

lib/intro_router.dart
```diff
@@ -1,32 +1,27 @@
...
+import 'package:wh_covid19/view/home_page.dart';
```
lukesleeman: I'm not a flutter dev, but this should fix up the flutter analyze warning

```suggestion
import 'view/home_page.dart';
```

lib/intro_router.dart
```diff
@@ -1,32 +1,27 @@
...
-    } else {
-      await Navigator.of(context).pushReplacementNamed(Routes.disclaimer);
-    }
+    return FutureBuilder<SharedPreferences>(
+      future: SharedPreferences.getInstance(),
+      builder: (context, snapshot) {
+        bool disclaimerFirstView;
+        if (snapshot.data != null) {
+          disclaimerFirstView =
+              snapshot.data.getBool("disclaimer_first_view") ?? false;
```
lukesleeman: I'm not a flutter dev, but this should fix up the flutter analyze warning

```suggestion
              snapshot.data.getBool('disclaimer_first_view') ?? false;
```

lukesleeman (commented):
@rockgecko-development I have tried to queue up the fixes for the flutter analyze in suggestions .  You should be able to just click on commit, commit, commit and merge.

PR #128, opened 2020-03-29
Fix card overflow, add touch feedback

rockgecko-development:
Fix overflow issue when text is too long for the fixed card height.
Issue Example (Society & College overflows):
image

Fixed by making `ReusableCard.height` the min height of the card, if supplied. It then supplies  `BoxConstraints(minHeight: height)` to ReusableCardBase. In case an explicit size is ever required in ReusableCardBase, this can be done with a different BoxConstraints.

Fix Example:
image

Also modified ReusableCardBase to use an `InkWell` for touch feedback.


lib/widget/cards/reusable_card_base.dart
```diff
@@ -59,7 +95,7 @@ class ReusableCardBase extends StatelessWidget {
         child: Padding(
           padding: padding,
           child: Container(
-            height: height,
+            height: null,
```
maks: isn't this the same as not specifying the height at all?

lib/widget/reusable_card.dart
```diff
@@ -15,7 +15,7 @@ class ReusableCard extends StatelessWidget {
   /// Route to view
   final String routeTo;
 
-  /// Height of card
+  /// Min height of card (default 84)
```
maks, line 18: :+1: :100:

lib/hard_data.dart
```diff
@@ -55,13 +55,11 @@ const List<ReusableCard> icu = [
   ),
   ReusableCard(
     title: 'General Care',
-    description: '',
```
maks, line 59: do we want to keep the description property with the empty placeholder for future use?

lib/widget/cards/reusable_card_base.dart
```diff
@@ -37,18 +37,54 @@ class ReusableCardBase extends StatelessWidget {
...
+                    child: col,
+                  ),
+          ),
+          onTap: () {
+            if (routeTo != null) {
+              Navigator.pushNamed(context, routeTo);
+            } else {
+              if (fallback != null) {
+                fallback();
+              }
```
maks: do we want to silently fail here? I think at least print a warning or maybe throw an Exception ?

maks (changes requested):
:tada: Great work @rockgecko-development ! Just a few small bits of feedback

lib/hard_data.dart
rockgecko-development: This is here because the vertical alignment of the text changes depending if description is null or ''. When description is null, we want title to be centred. https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/blob/8c70decee808c35c293093ddc300720f1d663ebe/lib/widget/reusable_card.dart#L52

lib/widget/cards/reusable_card_base.dart
rockgecko-development: We need to cater for cards that have no click action, for example `ppe_step_card.dart`.
But I'll make a quick change to mean `onTap` is null if there is no action, meaning there will be no touch feedback either, which is desirable.

lib/widget/cards/reusable_card_base.dart
rockgecko-development: Oops, this shouldn't be here at all, good pickup

maks (approved):
:tada: awesome work! :rocket:

sallychung:
@maks can you please kick code magic for this PR? It’s stuck in a state where we can’t merge this approved PR

maks:
@sallychung done: https://codemagic.io/app/5e75f4bfdd1d6d000a198e90/build/5e8312084b4d3278658308e8

rockgecko-development:
It failed because of this?
>info • Sort child properties last in widget instance creations • lib/widget/cards/reusable_card_base.dart:71:11 • sort_child_properties_last

This doesnt even show under the 'Dart Analysis' tab in Android Studio. Probably because it's at level `info`? This is so trivial, can we ignore this and move on?

Issue #112, opened 2020-03-28
Build out infographic screens


lukesleeman:
This issue depends on issue #111 and #96 being built out 1st.   Once we have our reusable infographic widget and the home screen is updated, there are 5 different infographics screens which need to be added to the app which will make use of it.

TASKS
- [x] PPE ON infographic
- [x] Taking off PPE Method One
- [x] Taking off PPE Method Two
- [x] Intubation Infographic
- [x] Intubation Checklist Infographic
- [x] Extubation Infographic

# PPE ON infographic
![Yourself : PPE : Detail 02 - Infographic](https://user-images.githubusercontent.com/902252/77846482-12da0300-7202-11ea-98aa-3c9d9e0000b1.png)

Image:  `putting_on_ppe.png`
left hand side button: `X`
Theme: Green
Title: `PPE ON Infographic`
How you access the screen: On the putting on PPE step by step guide screen add a button to view the PDF 👇

![image](https://user-images.githubusercontent.com/902252/77846494-2f763b00-7202-11ea-97e4-e6ee7b676f19.png)

# Taking off PPE Method One

![Yourself : PPE : Detail 04 - Method One Infographic](https://user-images.githubusercontent.com/902252/77846573-b3302780-7202-11ea-8811-b00faa27a89b.png)

Image:  `taking_off_ppe_1.png`
left hand side button: `X`
Theme: Purple
Title: `Method One Infographic`
How you access the screen: On the PPE screen, in the method one step by step guide  add a button to view the PDF 👇

![image](https://user-images.githubusercontent.com/902252/77846586-ccd16f00-7202-11ea-8987-ef15c105f218.png)

# Taking off PPE Method Two

![Yourself : PPE : Detail 06 - Method Two Infographic](https://user-images.githubusercontent.com/902252/77846596-dd81e500-7202-11ea-980b-743cb2e13342.png)

Image:  `taking_off_ppe_2.png`
left hand side button: `X`
Theme: Purple
Title: `Method Two Infographic`
How you access the screen: On the PPE screen, in the method two step by step guide  add a button to view the PDF 👇

![image](https://user-images.githubusercontent.com/902252/77846628-1ae67280-7203-11ea-894b-cd1ecee95735.png)

# Intubation Infographic

![Intubation : Guide - Infographic](https://user-images.githubusercontent.com/902252/77846675-6b5dd000-7203-11ea-854b-7b436f600025.png)

[...]

PR #130, opened 2020-03-29
[Feature] Build Reusable Widget for Infographics Screen

### LINKED ISSUE #111: Build re-usable infographic widget
There are a number of infographic screens in the app.  They all need to

- Display a single pinch to zoom image
- Have either a "x" or "<" button in the left hand corner
~- Have a share button in the top right which brings up the OS share sheet to share the image~
- Have configurable theming.  Either blue, purple or green
- Have a configurable title.

**These are the colours to use in green mode**
Yourself : PPE : Detail 02 - Infographic

**These are the colours to use in purple mode**
Yourself : PPE : Detail 05 - Infographic
(note the grey background)

**These are the colours to use in blue mode**
ICU : General Care
(Yes, I know the picture should have the share icon, not the phone.  You know what to do 😁 )

~I'm not sure if we can do different icons per platform, but if possible on Android we should use the 
`icon_share_android.svg` and on ios `icon_share_ios.svg`~

yimingc2:
Resolved issue #111
## Abstract
This PR will create a reusable infographics page with a zoomable image. This page will be used in multiple places.

## Changes

- Add photo_view dependency
- Add configurable (image, color and title) infographics page for reuse
- Use different share icons based on platform (Android/iOS)
- Apply close (X) icon to left navigation bar button

## Screenshots
zoom

## Things to note
- Not sure what infographic image should be used so I put Intubation Checklist Image for now
- The reusable infographics page has been applied to the Algorithm page for now. The route will be changed in #112


lib/view/reusable_infographic_page.dart
```diff
@@ -0,0 +1,45 @@
...
+    return Scaffold(
+      backgroundColor: backgroundColor,
+      appBar: AppBar(
+        iconTheme: AppStyles.appBarIconTheme,
+        title: Text(
+          title,
+          style: AppStyles.appBarTextStyle,
+          ),
+        actions: <Widget>[
+          IconButton(
```
maks: could we please break this out into a separate "platform icon button" widget, which takes a named const that represents the required icon, the constant can then be a map, with each asset path keyed by platform name

lib/widget/reusable_zoomable_widget.dart
```diff
@@ -0,0 +1,35 @@
...
+          crossAxisAlignment: CrossAxisAlignment.stretch,
+          children: <Widget>[
+            Expanded(
+              child: ClipRect(
+                child: PhotoView(
+                  backgroundDecoration: BoxDecoration(
+                    color: backgroundColor
+                  ),
+                  imageProvider: AssetImage(imageUrl),
+                  customSize: Size(MediaQuery.of(context).size.width - margin * 2, MediaQuery.of(context).size.height - margin * 2),
```
maks: could we break this calc into a separate function that returns the `Size` for clarity?

maks (changes requested):
:tada: excellent work @yimingc2 ! Just a few small issues to address.

lib/view/reusable_infographic_page.dart
yimingc2: Do you mean create a general reusable platform icon button or just the button for sharing? If it's first one, then we will need to pass the action as well.

maks (approved):
:+1: :100:

PR #135, opened 2020-03-30
Const explorations

domesticmouse:
Make more things `const`


maks (approved):
:+1: With the consts - were they not being picked up by the lint rule? as I have lint passing as mandatory for CI builds

domesticmouse:
@maks I think I am being more strict on the `const` front than the lint checks. I'll discuss a bit more as to why on thursday night =)

PR #138, opened 2020-03-31
Reduced file size of many PNG images.

marcedwards:
Many of the PNG images have been reduced in file size, using a variety of lossless and lossy techniques — I think it’s okay to use lossy PNG compression on the infographics, but I avoided it for the app icon and in places where smooth gradients would have become posterised.

I also tried some of the infographics as JPEGs, but the file size ended up being larger for comparable quality.

The end result:

Before = `3,742,993 bytes`
After = `1,986,392 bytes`

Difference: `1,756,601 bytes`

I’m not sure how much further things can be reduced without noticeable quality losses. We may also need more assets for the icons? Right now, only one size is shown, but iOS and Android require many more.


maks (approved):
:+1: excellent work @marcedwards !

lukesleeman:
I've run the app and verified the images are displaying well.  I'll merge this now

PR #149, opened 2020-04-02
Fix/disclaimer update

workerbee22:
## Changes

- Fix app startup to display the disclaimer on first run only or if the version has changed from the version the user accepted
- Fix back button so it cannot be used to skip the disclaimer screen
- Fix bug where home screen may be shown twice or Back button twice needed to exit the app on Android. iOS does not permit an app to exit itself.
- Fix bug where we get Overflow errors on some lower density device screens
- Add versioning within the app, so if the version flag persisted to device storage changes compared to the current app release, then reprompt user to accept the disclaimer
- New version of the Disclaimer page content
- Added scrollbar on touch when scrolling
- Added button fixed at bottom of the screen rather than have to scroll to bottom to press
- SafeArea for disclaimer so does not impact system UI on Android or iOS
- Information screen disclaimer shows that it was accepted by the user, version and date & time user agreed to the disclaimer

## Things to note

There are also some general app improvements like:
- strings.dart file for fixed strings used in code
- changes to some routes to fix the double back issue
- other general cleanup required.


workerbee22:
This isn't super tight code, but it is quite urgent (apologies been doing CI & CD setup). So keen to merge this and then do a separate PR for making it look nicer.

workerbee22:
Ok just pushed a new commit that has caught up with master branch. 

Note many file changes are due only to:
- Strings class in `strings.dart` file
- Styles class renamed from AppStyles previously

sallychung:
@workerbee22 I noticed that the if I access the Disclaimer from the Info section, there’s no back button so I get stuck. Would you like me to raise an issue for this as a separate task?

iPhone 6s
image

workerbee22:
@sallychung Yep its a bug.

workerbee22:
So I'm working on these before finalise this PR:

- fixing the bug about no back button on ios
- fixing tests for updated shared preferences values and adding new tests

lib/view/disclaimer_view.dart
```diff
@@ -1,206 +1,202 @@
...
 }
 
-class _DisclaimerViewState extends State<DisclaimerView>
-    with SingleTickerProviderStateMixin {
-  final String _title = 'Disclaimer and conditions of use';
+class _DisclaimerViewState extends State<DisclaimerView> {
+  @override
+  void initState() {
+    super.initState();
+  }
```
alessandrofavero: If you are not doing anything in the initState(), I would remove it.

lib/view/disclaimer_view.dart
```diff
@@ -1,206 +1,202 @@
...
-    child: const Text(
-      disclaimerBody,
-      style: AppStyles.textLegal,
-      //textAlign: TextAlign.justify,
-    ),
-  );
+  @override
+  void dispose() {
+    super.dispose();
+  }
```
alessandrofavero: Same as per initState(), more compact code.

lib/view/disclaimer_view.dart
```diff
@@ -1,206 +1,202 @@
...
+    return disclaimerValues;
   }
 
   @override
   Widget build(BuildContext context) {
-    // Bottom button to agree t&c
-    final _agreeButton = Container(
-      margin: const EdgeInsets.only(left: 16, right: 16, bottom: 16),
+    /// Agree button - button a user taps to agree to the disclaimer
+    final Widget _agreeButton = Container(
```
alessandrofavero: I would convert this to a function like _agreeButton(BuildContext context) and move it outside the build method.

lib/view/disclaimer_view.dart
```diff
@@ -1,206 +1,202 @@
...
-                            child: ListView(
-                              children: <Widget>[
-                                _content,
-                                if (!snapshot.data)
-                                  _agreeButton
-                                else
-                                  _agreedText,
-                              ],
+    // Content of the page
+    Widget _agreedMessage(dynamic version, dynamic dateStampString) {
```
alessandrofavero: I would move this function outside the build method

lib/view/disclaimer_view.dart
```diff
@@ -1,206 +1,202 @@
...
+      return Container(
+        padding: const EdgeInsets.all(16),
+        child: Text(
+          '${Strings.disclaimerHaveAgreedText} \nVersion: $version \nDate & time: $dateStampString',
+          style: Styles.textLegal,
+        ),
+      );
+    }
+
+    Widget _disclaimerScrollingContent(BuildContext context) {
```
alessandrofavero: I would move this function outside the build method

lib/view/disclaimer_view.dart
```diff
@@ -1,206 +1,202 @@
...
-    await prefs.setBool('disclaimer_first_view', true);
+    return SafeArea(
+      bottom: true,
+      top: true,
+      child: Scaffold(
+        appBar: AppBar(
+          backgroundColor: Colors.white,
+          elevation: 4.0,
+          iconTheme: Styles.appBarIconTheme,
+          automaticallyImplyLeading: false,
```
alessandrofavero: You need to implement a check here otherwise the user will not be able to go back if they check the disclaimer from the info page.

alessandrofavero (commented):
Hey @workerbee22 , there are some minor changes that I would address and a check implementation in the automaticallyImplyLeading. Other than that LGTM great job 👍👍

lib/view/disclaimer_view.dart
workerbee22: Good idea 👍

lib/view/disclaimer_view.dart
workerbee22: Yeah thats the bug in the Information screen view. Need to move the FutureBuilder up to get that value in here.

lib/view/disclaimer_view.dart
workerbee22: Good point ... I was playing with something in there. I usually keep initState and dispose in there regardless .... but thats just me. I'll remove 👍

lib/view/disclaimer_view.dart
workerbee22: ditto 👍

lib/view/disclaimer_view.dart
workerbee22: Good idea 👍

lib/view/disclaimer_view.dart
workerbee22: ditto 👍

lib/view/disclaimer_view.dart
workerbee22: I'm still torn on the Disclaimer view at app startup vs Disclaimer view from the Information screen, because the only thing they really have in common is the disclaimer content itself. But happy to keep your original approach of using it for both for now, unless they change more in the future and get even more different 👍

workerbee22:
Thanks @alessandrofavero ... good ideas. I also wanted to keep your original approach as much as possible. Back to desk now and working on these.

Also have some tests to fix and additional tests, but want this in in next few hours 👍

lib/view/disclaimer_view.dart
workerbee22: All done.

workerbee22:
@alessandrofavero 
All raised issues resolved and changes made, can you perhaps review if you get a chance because you know this code best. Then I can merged.

workerbee22:
@maks @ericjiang97 @rockgecko-development If one of you could also review during the day? We can finally get this in ... pretty later.

lib/view/disclaimer_view.dart
```diff
@@ -37,6 +28,7 @@ class _DisclaimerViewState extends State<DisclaimerView> {
     final disclaimerValues = DisclaimerDetails();
 
     disclaimerValues.agreed = prefs.getBool(Strings.settingDisclaimerAgreed) ?? false;
+    // Disclaimer version flag starts at '1' normally
     disclaimerValues.version = prefs.getString(Strings.settingDisclaimerVersion) ?? '0';
```
alessandrofavero: cascade?
final disclaimerValues = DisclaimerDetails()
..agreed = .......
..version = ....... ;

alessandrofavero (approved):
LGTM 👍Tested the codemagic build on Pixel4XL and it works great.

workerbee22:
@alessandrofavero Wrote the original code and knows the features around Disclaimer really well. I've also tested on 4+ devices iOS and Android, so going to merge it.

lib/view/disclaimer_view.dart
workerbee22: Given Disclaimer is all tied up with app startup behaviour, I thought I'd make the code a little more 'readable' vs less lines of code. We can definitely tighten up, but with many new Flutter devs on the project I'm err-ing on the side of 'readability' an making the code clean and simple.

workerbee22: But we should discuss this sort of stuff in the nightly meetings or slack. Hard to know what everyone would prefer?

workerbee22:
All good ... merged into master, built, published and tested on Android & iOS 👍

PR #151, opened 2020-04-03
Fixed disclaimer tests x 3 

workerbee22:
Fixed the Disclaimer widget tests for app startup behaviour (thanks to @maks for a fix).

## Changes

Widget tests for disclaimer are now: 

- Tests for correct app startup screen either home or disclaimer
- Disclaimer shown if user has never agreed to disclaimer or version flag has changed since they approved previously
- Home shown if they have agreed to the disclaimer and version flag matches the current version of the disclaimer in the app

## Things to note

Tests now use Strings class string values, rather than hard coded text in the test. This should make code to test mismatches in text less likely. For example a title could change in Strings class and this would be reflected in the UI and also used by the tests.

Home screen code updated to use Strings class value for the UI code.

### DIFF
--- a/lib/strings.dart
+++ b/lib/strings.dart
@@ -19,7 +19,12 @@ This app is designed to be used by Anaesthetic and Critical Care doctors at West
   // *** WARNING ***
   // Increment by one when the disclaimerBody changes. This will force users to accept/re-accept the disclaimer again
   // if they have previously accepted it. If the disclaimer has changed then users need to re-accept it.
-  static const String disclaimerVersion = '1';
+  static const String disclaimerCurrentVersion = '1';
   static const String disclaimerButtonAgreeText = 'I Agree';
   static const String disclaimerHaveAgreedText = 'You have agreed to the disclaimer';
+
+  /// Home strings
+  static const String homeHeading1 = 'Look After Yourself';
+  static const String homeHeading2 = 'Airway';
+  static const String homeHeading3 = 'ICU';
 }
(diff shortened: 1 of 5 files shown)


lib/strings.dart
```diff
@@ -19,7 +19,12 @@ This app is designed to be used by Anaesthetic and Critical Care doctors at West
   // *** WARNING ***
   // Increment by one when the disclaimerBody changes. This will force users to accept/re-accept the disclaimer again
   // if they have previously accepted it. If the disclaimer has changed then users need to re-accept it.
-  static const String disclaimerVersion = '1';
+  static const String disclaimerCurrentVersion = '1';
```
maks, line 22: Actually thinking about it, might be better to make this a number type so can do '>' comparison on it

maks (approved):
👍 just a small suggestion

lib/strings.dart
workerbee22: Actually I intentionally made it String. Because its only important that it be different, not incrementing. I think its one of those rare cases where ... handy to be a number because humans relate well to that + only important that we detect a change (which we'll be doing manually).

workerbee22: Also we may have to roll back a version of the Disclaimer, which if a number then makes it interesting.

workerbee22: Each release code only holds the current version flag, so only important to detect device version flag vs current app version flag. The app can't know previous values in previous app releases. .... I thought about this one a lot before making it a number as String.

ericjiang97 (approved):
Looks good.

lib/strings.dart
workerbee22: @maks I can change all the code to make it an int ... but it won't get us anything. behaviour will be exactly the same.

maks (approved):
👍

PR #162, opened 2020-04-04
Update Pod lock file

sallychung:
Update the Pod lock file to use the latest version of Cocoapods


workerbee22 (approved):
Tested on iPhone 6 run local. `pod update` runs fine. `pod install` for flutter run, works fine.

LGTM

PR #168, opened 2020-04-04
Fixed layout on home page 

alessandrofavero:
- [x] Switched CardContainer layout on enums instead of cards.lenght
- [x] Implemented a new three cards column layout

ezgif com-resize


workerbee22 (approved):
LGTM

workerbee22:
@alessandrofavero Approved, although we'll need to talk about perhaps LOGGING in the app, which can be later. @maks I'll add a issue for app logging setup and another for code patterns using `assert`.

Also probably prefer `assert` statements over this approach, but also needs to be code wide:

image

Issue #147, opened 2020-04-01
Build a Licenses screen


lukesleeman:
We will probably need to add in a licenses screen for the various bits of open source software we are including

lukesleeman:
@marcedwards or @Mattro when you get some time, can you put a button to link out to licenses on the information screen.

In terms of how the licenses screen its self needs to look - I think just a huge wall of text?

workerbee22:
Note Flutter has two classes for these sorts of things. But they result in a default UI. We can roll our own but use these to automatically provide the info for the app, especially all licences info:

https://api.flutter.dev/flutter/material/LicensePage-class.html
https://api.flutter.dev/flutter/material/AboutDialog-class.html

It allows for easily getting the current app icon, version and other details. So we should use these classes where appropriate for the data from the app. No hard coding data !

So we just need the 'look' designed.

lukesleeman:
Yeah, I'm thinking for the licenses screen, there is no point spending any design time on it.  I think if @Mattro or @marcedwards can just add a button to the information screen to launch it, and update the flow diagram, then we can move this one over to ready for dev.

maks:
:+1: to using the builtin AboutDialog and LicensePage we get alot of functionality for free and no need to spend effort on this non-core functionality.
@lukesleeman can this go out for dev to pickup now?

lukesleeman:
> 👍 to using the builtin AboutDialog and LicensePage we get alot of functionality for free and no need to spend effort on this non-core functionality.
> @lukesleeman can this go out for dev to pickup now?

@maks any idea on how we go about getting a list of licenses we need to include?

maks:
@lukesleeman The beauty of it is it should all be automatic with the package manager taking care of it: https://api.flutter.dev/flutter/foundation/LicenseRegistry-class.html

lukesleeman:
Oh, nice!  In that case moving this one to ready for dev!

lukesleeman:
Thanks for the updates to the diagram @cjmlgrto  - I'm moving this one over to dev 🙂

PR #172, opened 2020-04-04
Replace Information Screen's "Lorem Ipsum" with proper copy

### LINKED ISSUE #166: Update Information screen & remove Lorem ipsum
Simulator Screen Shot - iPhone 5s - 2020-04-04 at 18 12 28

cjmlgrto:
This PR will resolve #166 by replacing the “Lorem Ipsum” copy in the app’s Information Screen with more appropriate copy about the app.

## Changes

Updates the Information Screen (`info_view.dart`) to use the following copy, sourced from the repo's README intro paragraph:

> This app aims to help safeguard the welfare of Western Health anaesthetists working amidst the COVID-19 pandemic by providing the most up-to-date information for time-critical procedures to ensure optimal patient care.

## Screenshots

Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-04 at 21 31 02

### DIFF
--- a/lib/view/info_view.dart
+++ b/lib/view/info_view.dart
@@ -106,7 +106,7 @@ class InfoView extends StatelessWidget {
       const Padding(
         padding: EdgeInsets.all(8.0),
         child: Text(
-            'About this app. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Purus gravida quis blandit turpis cursus in hac habitasse platea. In vitae turpis massa sed elementum tempus. Pulvinar neque laoreet suspendisse interdum consectetur libero.'),
+            'This app aims to help safeguard the welfare of Western Health anaesthetists working amidst the COVID-19 pandemic by providing the most up-to-date information for time-critical procedures to ensure optimal patient care.'),
       ),
     ];
   }


workerbee22:
@cjmlgrto I think we should get Gregg to review/tweak this copy as the business owner, which is what we are doing with all the rest. 

I'm also not convinced this copy even needs to be here. Do we need it? What does it add? By the time the user is in here, they have seen the app and get a feel for what it is. If anything key messages here would be in onboarding screens, but we've de-prioritised that for now.

workerbee22:
Another related issue is here : https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/147

Its related because things like version are super easy to get from some Flutter apis listed in that issue too. 👍

cjmlgrto:
@workerbee22 good point — best to drop the paragraph in the first place then 👍 

Should I close this PR and make a new one that suggests dropping it?

workerbee22:
Lets keep it open and get some more comments Im thinking?

kimngyn:
It looks very nice! I think the hospital staff would rather read this than Lorem Ipsum. I’m not very good with Latin 🤗

greggmiller:
Thanks for putting the work into this @cjmlgrto 😄 , I think Mike's idea is good that we can remove this and close it down. All the information is accessible elsewhere and studies up this screen a bit if we remove it 👍

PR #182, opened 2020-04-05
apply dartfmt across entire codebase

maks:
To stop needless churn in dart source code from formatting changes, dartfmt has been applied ot all dart files and developers will now need to do so before merging to master as the CI PR-workflow will now check for correct src formatting and fail if formatting has not been applied.


lukesleeman (approved):
Looks very ✨ - good work @maks

PR #184, opened 2020-04-05
Operating system support updates

### LINKED ISSUE #176: Support only iOS 12 and above
Emojis aren’t displayed correctly on iOS 11.0.1
- Eat well
- Stay in contact with loved ones
- Avoid unhelpful coping strategies...

Note: This was tested on simulator only as I don’t have a physical device

Simulator Screen Shot - iPhone 5s - 2020-04-04 at 22 12 46

### LINKED ISSUE #177: System Bars styling
The status bar on iOS and Android and the navigation bar on Android are usually styled by the system. For example by ensuring contracts and color, icons can always be clearly visible for users.

- **Operating System version variations** - Different versions of iOS and Android take different approaches to ensuring clarity of system and navigation bars, and so the result varies based on the operating system version. On Android different launchers also change this behaviour.

- **App Styling** - Apps can also change the system bar styling for their screens. Flutter automatically tries to set the apps system bar colors and style, based on the colors of certain widgets like AppBars. Apps also change the styling of system bars to re-enforce their brand design.

Currently in our app we have some issues on several screens where typically the status bar icons are not clear:

**iPhones 6 on iOS v12.x**

White status bar icons on white background:

image

Black status bar icons on black background:

image

**iPhone 11 Pro Max on iOS v13.3 (simulator)**

Black status bar icons on black background:

image

White status bar icons on white background when scrolling Home screen:

image

White status bar icons on white background 

image

### LINKED ISSUE #183: Lock android versions down to api 21 and above
The app needs to be changed to only support android versions on API 21 and above

The app currently supports some really old versions of Android.  As we are using a lot of emoji this can be a problem (see issue #176 for the problems on the iOS side).  Locking the app to API 21 won't fully fix the issues, however it will mean that we don't have to worry about some of the very oldest devices.  We discussed the issues on this evenings hangouts call and decided that going API 21 and above would be reasonable.

workerbee22:
Operating system related fixes and adjustments for iOS and Android.

## Changes

-Initial Themeing setup within the app.
-Status & Nav bar coloring fixed for Android and iOS.
-iOS support now limited to v12.0 and above.
-Android support now limited to v5.0 (API 21) and above.
-WHAC19 as the new product icon name on Android and iOS.
-App name changed in the app and other code optimisations.

## Things to note
Retested on iOS 12.x 13.x and Android 8.x and 9.x.


lib/view/home_page.dart
```diff
@@ -128,10 +132,17 @@ class _HomePageState extends State<HomePage> {
 
   @override
   Widget build(BuildContext context) {
```
kimngyn, line 134: nice!

workerbee22:
@rockgecko-development @sallychung and @maks ... hoping you can look at this one soon. Lots of os changes and some nice cosmetic ones too.

analysis_options.yaml
```diff
@@ -11,7 +11,7 @@ linter:
     - avoid_bool_literals_in_conditional_expressions
     - avoid_catches_without_on_clauses
     - avoid_catching_errors
-    - avoid_classes_with_only_static_members
+#    - avoid_classes_with_only_static_members
```
maks: to be explicit for future readers pls set to false rather than commenting out

lib/main.dart
```diff
@@ -118,6 +115,7 @@ class MyApp extends StatelessWidget {
     );
   }
 
+  // TO DO - This can be optimised probably to remove html views.
```
maks: better to raise this in an issue vs using a TODO comment

lib/style.dart
```diff
@@ -54,8 +54,18 @@ abstract class AppColors {
 }
 
 // TODO: Convert members to top level finals/consts or functions.
+// ignore: avoid_classes_with_only_static_members
```
maks: this ignore is not needed with change in the yaml config

lib/utils/system_bars.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+/// System bars styling is set
```
maks, line 4: should be consistent and use `///` for the whole comment

lib/utils/system_bars.dart
```diff
@@ -0,0 +1,27 @@
...
+//
+// If you have an AppBar or SliverAppBar then it will interfere with your
+// calls to SystemChrome.setSystemUIOverlayStyle and you can instead just
+// change the brightness property on your AppBar.
+SystemUiOverlayStyle systemBarStyle(BuildContext context) {
+  // Copy the default style and then modify key values
+  final lightStyle1 = SystemUiOverlayStyle.dark.copyWith(
+    // Status bar - iOS and Android
+    statusBarBrightness: Brightness.light, // iOS only
+    statusBarColor: Colors.transparent, // Android v6 and above
```
maks: what does `Android v6 and above` mean when we only support SDK21 and above?

lib/view/icu_non_intensivist/general_care_view.dart
```diff
@@ -11,9 +11,13 @@ class GeneralCareView extends StatelessWidget {
     return Scaffold(
       backgroundColor: AppColors.backgroundBlue,
       appBar: AppBar(
-          backgroundColor: AppColors.appBarBackground,
-          iconTheme: Styles.appBarIconTheme,
-          title: Text(title, style: Styles.appBarTextStyle)),
+        // Warning brightness interacts with SystemUiOverlayStyle
+        // See system_bars.dart comments
+        brightness: Brightness.light,
```
maks, line 16: do we need to hardcode this, wont it default to `.light` ?

lib/view/view_templates/html_text_card_view_template.dart
```diff
@@ -13,17 +13,26 @@ class HtmlTextCardViewTemplate extends StatelessWidget {
...
-            title: Text(title, style: Styles.appBarTextStyle)),
-        body: SingleChildScrollView(
-            child: Padding(
-                padding: const EdgeInsets.all(12),
-                child: Card(
-                    child: HtmlWidget(html,
-                        textStyle: Styles.cardDescriptionTextStyle,
-                        bodyPadding: const EdgeInsets.all(6))))));
+      backgroundColor: bgColor ?? AppColors.backgroundBlue,
+      appBar: AppBar(
```
maks, line 17: Seem to repeat this Appbar config several diff places, can we just have a builder function or custom widget for it t keep it more DRY ?

lib/view/view_templates/html_text_card_view_template.dart
```diff
@@ -13,17 +13,26 @@ class HtmlTextCardViewTemplate extends StatelessWidget {
...
-            iconTheme: Styles.appBarIconTheme,
-            title: Text(title, style: Styles.appBarTextStyle)),
-        body: SingleChildScrollView(
-            child: Padding(
-                padding: const EdgeInsets.all(12),
-                child: Card(
-                    child: HtmlWidget(html,
-                        textStyle: Styles.cardDescriptionTextStyle,
-                        bodyPadding: const EdgeInsets.all(6))))));
+      backgroundColor: bgColor ?? AppColors.backgroundBlue,
```
maks, line 16: maybe nicer to just set the default value on the bgColor param in constructor

maks (changes requested):
:+1: great work @workerbee22 just some small bits and pieces.

analysis_options.yaml
workerbee22: Oh yeah forgot to go back and do that, good pickup 👍

lib/main.dart
workerbee22: Good point.

lib/style.dart
workerbee22: Yep cool. That actually wasn't me, but will remove 👍

lib/view/icu_non_intensivist/general_care_view.dart
workerbee22: I'm afraid so for now. Not sure why just yet, but AppBar not following the theme properly (which has Brightness.light) and so without it we get the white on white status bar again on any AppBar that does not have it. So have to do it on every AppBar for now.

I suspect this is a bug in current Flutter version because works fine and picks up the theme brightness in my other apps that are on older Flutter `stable`.

lib/view/view_templates/html_text_card_view_template.dart
workerbee22: Yep afraid no way around it right now.

lib/view/view_templates/html_text_card_view_template.dart
workerbee22: Will look at that, but yep one of manu optimisations we still need to make due to lots and lots of duplication. But one step at a time.

analysis_options.yaml
workerbee22: @maks any idea on how to disable lint rules. Looks like the syntax here in doco is WRONG: https://dart.dev/guides/language/analysis-options#disabling-individual-rules

Throws Dart Analysis errors.

lib/style.dart
workerbee22: So need to confirm how to disable before i can remove this, otherwise lint rule gets triggered still.

lib/utils/system_bars.dart
workerbee22: It means this property is only effective on Android 6 and above. So doco exact quote is:

'Only honored in Android version M and greater'

Added these comments otherwise anyone reading the code will get very confused very fast as some properties iOS only, some only specific Android versions. I'll add comments above the line and quote dodo.

workerbee22: Its a pain because they are almost all different. So have added better comments above lines. 👍

analysis_options.yaml
maks: `Note: Due to YAML restrictions, you can’t mix list and key-value syntax in the same rules entry`

maks: So yeah unfortunately need to either just delete the line with that lint or change all the other rules to use key:clue syntax 😟

lib/style.dart
workerbee22: Also I don't think we want to do these TODOs that were here already:

// TODO: Convert members to top level finals/consts or functions.

Thoughts @maks ?

analysis_options.yaml
workerbee22: There seem to be many grumbles on the net about this.

lib/utils/system_bars.dart
maks: Ah ok in that case better to use Android version letters or even better the SDK number, for Android devs bit confusing to see Android version numbers not letter names.

lib/view/view_templates/html_text_card_view_template.dart
maks: But we could have a single function that returns this appbar widget

analysis_options.yaml
workerbee22: OK I converted to the other format and then disabled this lint rule. 👍

lib/style.dart
workerbee22: Now that analysis_options.yaml fixed. Removed this line.👍

lib/utils/system_bars.dart
workerbee22: Have letter and API level now in comments 👍

lib/utils/system_bars.dart
workerbee22: Using all /// for whole comment section 👍

lib/view/view_templates/html_text_card_view_template.dart
workerbee22: @maks Yes we could and should, but would prefer to do that in a different wider PR if thats OK. We have too many AppBar widgets anyway. We have a couple of key screen layouts that can use a new custom widget, that contains the AppBar. 

So thats a bigger chunk of work, so prefer to keep this to the scope of just OS fixups. This PR is getting a bit big as it is.

workerbee22:
@sallychung review mostly done with @maks on Flutter changes and Android, but will wait for your review on iOS side.

ios/Podfile.lock
```diff
@@ -106,4 +106,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: c34e2287a9ccaa606aeceab922830efb9a6ff69a
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.8.4
```
yimingc2: Any reason we downgrade cocoapods version? Maybe running locally changed it by accident.

yimingc2 (changes requested):

maks (approved):
👍

ios/Podfile.lock
workerbee22: @yimingc2 I'll do a pod update and re push.

yimingc2 (commented):
Amazing work!

ios/Podfile.lock
yimingc2: @workerbee22 I think your local cocoapods version is 1.8.4. At the moment, we don't have Gemfile to let everyone run the same cocoapods version. So I think we can simply ignore it for now. If we don't need to update any dependencies, then we don't need to `pod update`. If someone just downloads the app and start from stretch, he can use `pod install` to download the required version specified in Podfile.lock

rockgecko-development:
I checked it out, ran it on Android, LGTM, thanks @workerbee22 !

yimingc2 (approved):

ios/Podfile.lock
workerbee22: Fixed with last commit. Now 1.9.1 👍

PR #186, opened 2020-04-05
Fix issue #171

### LINKED ISSUE #171: Intubation Checklist resets when traversing between tabs
**Steps to reproduce**
1. Go to Intubation Checklist 
2. Select cells under Teams
3. Go to any other tab and select cells, ie, Patient
4. Go back to Teams

Expected:
Selected cells should remain selected

Actual:
All cells have been deselected

checklist_bug

rockgecko-development:
Fix issue #171 by storing checked state in ChangeNotifierProvider class in IntubationChecklistPage

### DIFF
--- a/lib/widget/lists/intubation_checklist_item.dart
+++ b/lib/widget/lists/intubation_checklist_item.dart
@@ -8,7 +8,7 @@ import '../../widget/cards/reusable_card_base.dart';
 import '../../widget/checkbox.dart';
 import '../../widget/lists/string_list.dart';
 
-class IntubationChecklistItemWidget extends StatefulWidget {
+class IntubationChecklistItemWidget extends StatelessWidget {
   final IntubationChecklistItem listItem;
   final Color backgroundColor;
   final Color selectedBackgroundColor;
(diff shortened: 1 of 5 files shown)


workerbee22:
@rockgecko-development Looking good and this essentially sets our state management solution to Provider. Which I'm fine with. @maks you happy with Provider? BLOC seems way to heavy for this app.

maks:
I actually use both Bloc & provider. I'm a big fan of provider, so 💯 ok with using it.

lib/widget/lists/intubation_checklist_item.dart
```diff
@@ -22,7 +24,7 @@ class IntubationChecklistItemWidget extends StatefulWidget {
 
 class _IntubationChecklistItemWidgetState
     extends State<IntubationChecklistItemWidget> {
-  var _checked = false;
+  //var _checked = false;
```
maks: delete instead of commenting out?

lib/widget/lists/intubation_checklist_item.dart
```diff
@@ -35,25 +37,27 @@ class _IntubationChecklistItemWidgetState
...
+        Provider.of<IntubationChecklistSelectionProvider>(context);
+    final isChecked = selectionState.isChecked(widget.listItem);
     return ReusableCardBase(
       elevation: 0,
       fallback: () {
-        setState(() {
-          _checked = !_checked;
-        });
+        // setState not needed, because Provider takes care of rebuilding.
+        selectionState.setChecked(widget.listItem, checked: !isChecked);
```
maks: can the item widget be stateless now too?

maks (changes requested):
💯 👍 on switching to use provider for state @rockgecko-development

lib/widget/lists/intubation_checklist_item.dart
rockgecko-development: ah forgot to clean that up

maks (approved):
:100: :+1:

Issue #197, opened 2020-04-07
Use of Strings class in strings.dart


workerbee22:
Already partially implemented, but this issue about cleaning up the code to FULLY support the use of a strings file.

### Proposal

Instead of embedding or 'hard coding' string constants in ui code, use a Strings class member instead. The benefits are 

- maintainability as key strings can then be easily changed in one place without touching the ui code itself.
- more readable code

For example for the app name `WHAC19` reference this in ui code as

```
return MaterialApp(
   title: Strings.appName,
```
The `Strings` class defined in `strings.dart` for example the current file:
![image](https://user-images.githubusercontent.com/5154664/78626202-666ae180-78d1-11ea-9ea9-fd09461a05b7.png)

### When to use Strings vs Content Data

In this WHAC19 app the 'data' is essentially the content displayed to the  app users. The content data we use in the app given the pandemic, approvals etc. may be MORE likely to change than Strings. Currently we have this data content embedded into widgets used either:

- in directly embedded in the ui code like this:

![image](https://user-images.githubusercontent.com/5154664/78625749-438bfd80-78d0-11ea-97bd-9afc81ecfff4.png)

- widgets referenced from `hard_data.dart` like this:

![image](https://user-images.githubusercontent.com/5154664/78625830-69190700-78d0-11ea-962d-4cb88236a572.png)

**Propose that the use of Strings does NOT impact on the content data, which is a separate issue. See background in decision [ADR-0002 ](https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/blob/master/docs/Architecture/ADR-0002%20Embed%20content%20in%20the%20app.md)**

Suggest using Strings references when:
- semi permanent ui elements like eg. AppBar titles, menu item text, app name etc.
- this may apply to the data in `hard_data.dart` as well
- but this is not a black &  white rule, just a guide.

lukesleeman:
These sound like some sensible suggestions 👍

joshcase:
Great idea. Is this up for grabs @workerbee22 or are you half way down the rabbit hole?

workerbee22:
Almost finished @joshcase . See this PR https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/pull/207

So this one nearly done. But check issues list, still many open.

workerbee22:
Finished today.

sallychung:
@workerbee22 would this require checking the text for every single screen?

[...]

PR #207, opened 2020-04-08
Code cleanup use Strings and fix Trailing Commas

workerbee22:
This is a code cleanup and consistency change. 

- See issue [197](https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/197) for the recommendation on the Strings class.
- trailing commas code formatting and alignment as per recommended approach [Flutter Trailing Commas](https://flutter.dev/docs/development/tools/formatting#using-trailing-commas)

## Changes

- String class use for on screen text, except for content/data
- `constants.dart` for app constants that should never change
- code reformatting to add trailing commas as per recommendation above

## Screenshots

No UI changes as this is a code cleanup exercise only. App tested and re run on Android and iOS to confirm no changes.

Note comparisons to Zeplin designs are yet to be done due to the time involved and the urgency on merging this during the day before other work on the code base starts later tonight. This change will make alignment to design easier.

## Things to note

Most files in the project have changed and so this is a very large PR. Please resist other changes in this PR, we can do it in others later.

Note there are still some future changes to `hard_data.dart` to use Strings more often that should be done, but plan to do that in another PR (given its just those two files involved) given the line between Strings and content/dart is unclear here.

### DIFF
--- a/lib/hard_data.dart
+++ b/lib/hard_data.dart
@@ -6,11 +6,18 @@ import 'models/IntubationChecklist.dart';
 import 'models/PPEStepInfo.dart';
 import 'models/intubation_guide.dart';
 import 'models/ventilation_guide.dart';
-
 import 'routes.dart';
 import 'strings.dart';
 import 'style.dart';
 import 'widget/reusable_card.dart';
+
+/// Web urls
+const String whURL = 'http://wh.cyphix.net/';
+const String whCoronavirusInfoURL = 'https://coronavirus.wh.org.au/';
+const String whAnaestheticMicrositeURL =
+    'https://whcovid19.wixsite.com/covid19';
+const String whIcuMicrositeURL = 'https://whicu2020.wixsite.com/icuquickguide';
+const String feedbackFormUrl = 'https://forms.gle/zQtfhvswrKmjJjNV7';
 
 // Staff Welfare card list composition
 const List<ReusableCard> staffWelfare = [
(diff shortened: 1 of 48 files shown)


domesticmouse (approved):

ericjiang97 (approved):

lib/hard_data.dart
```diff
@@ -21,7 +28,7 @@ const List<ReusableCard> staffWelfare = [
     routeTo: Routes.ppe,
   ),
   ReusableCard(
-    title: 'Your Welfare',
+    title: Strings.yourWelfareTitle,
     description: 'Vital info & contacts',
```
rockgecko-development, line 32: I wonder if it makes sense to extract the titles, but not the descriptions? I think it's fine with either both in strings, or both here.

rockgecko-development:
LGTM :shipit:

lib/hard_data.dart
workerbee22: Yeah .... was up and down on that too. In `hard_data.dart` there are lots of these choices.

In this case the title is used multiple places in code so makes sense ie. 'Your Welfare' in the cared title on the Home screen, but also on the AppBar within that screen. So makes sense to keep those two in sync.

I left the description as only used once. So either would be fine I guess. But for example for ALL the strings in `hard_data.dart` I don't think we want to move them to Strings, as almost no advantage vs leave where they are. 

So deferred these choices for now. But at least its just these 2 files involved. So much easier to chnage this later.

rockgecko-development (approved):

workerbee22:
OK with codemagic down, have done all local tests like `flutter analyze`  and `flutter test` and all passed. Also tested and found a bug, so a final commit to fix.

Now merging to master manually, skilling codemakic PR workflow build. So we can get this one done.

PR #228, opened 2020-04-10
Fix for App Store Connect 'ITMS-90381: Too many symbol files'

workerbee22:
This is an Xcode config change only to clean up this App Store Connect warning, which only presents when App Store Connect complains about it. Needed to get ready for wider release.

## Changes

- Xcode configuration change to remove the warning 'ITMS-90381: Too many symbol files'

## Things to note

- No ui changes to the app
- No update to release notes as is not user facing


workerbee22:
PR workflow passed and this is a small non-ui change and need to force master update and push to App Store Connect before can confirm it fixes the issue. So pushing to master.

sallychung:
No QA required.

PR #232, opened 2020-04-10
Fixed Home Page Layout

### LINKED ISSUE #173: Main Screen - UI Inconsistencies
- [ ] A - Info button is not aligned to the title
- [ ] B - Content below title is not aligned on the left
- [ ] C - Buttons’s height under Intubation are too tall

Simulator Screen Shot - iPhone 5s - 2020-04-04 at 21 34 35

alessandrofavero:
Fixed #173

- [x] A - Aligned info button to the top of the logo
- [x] B - Content all aligned
- [x] C - Reduced height of the buttons

## Changes

- Dot points explaining what is changed
- Did the thing
- Changed some stuff

Screenshot_1586484279


lib/hard_data.dart
```diff
@@ -28,12 +28,14 @@ const List<ReusableCard> staffWelfare = [
     description: '3 guides',
     color: Colors.white,
     routeTo: Routes.ppe,
+    margin: EdgeInsets.all(0.0),
```
maks, line 31: Sorry I don't think putting layout/ui into the data is a good idea. A better approach would be to give these a diff type (eg make a subclass Reuseablecard) and then in UI apply the different margins/heights based on the data type

maks (changes requested):

lib/hard_data.dart
alessandrofavero: That's true, so then even the colour should not be there.  I went for a quick fix without impacting any other view that uses the ReusableCard class. I'll follow your approach.

lukesleeman (approved):
I've run the app on a device and confirmed that this does address all the tweaks in issue #173 

@maks makes some good points about how the code is structured.  But once they have been addressed, I'm happy for you to merge 👍

lib/hard_data.dart
rockgecko-development: Gonna have to disagree with @maks here. `ReusableCard` is a **widget**, not a data structure. It already contains fields for `height, elevation, color, fallback` onClick functions.
Because it already contained the color field, there I see no issue in assigning it here. Certainly, that makes more sense to me than creating new Blue, Green, and White Widget classes whose only difference is the colour.

@maks is trying to make it a data class. Ideally, it should be a pure data object, but that would have significant impacts, in particular for how the `fallback` handler is wired up.

maks: @rockgecko-development yes I only noticed that after I talked about this directly with @alessandrofavero on slack and had look at code in editor vs the diffs here.
So definitely this needs to be fixed, we have proper data model classes now so there should not be any Widgets in the data layer.
As discusses on slack with @alessandrofavero I'm fine for this PR to go in just fixing the immediate layout tweaks and Im going to follow up with PR to fix the structural issues.

maks (approved):

sallychung:
✅ v1.0.131 ✨ 
IMG_980B502018EB-1

PR #236, opened 2020-04-10
Feature - Firebase Analytics & Crashlytics

### LINKED ISSUE #15: Need to add tool for Crash Reporting & Analytics
I'm familiar with Sentry and it works well for Flutter in prev project so I can set up Sentry account and add to app unless someone else has different preferences?
cc: @workerbee22 @alessandrofavero @adamkoch @domesticmouse

workerbee22:
Initial implementation of Firebase Analytics (data is also passed to Google Analytics) and Firebase Crashlytics for crash & error reporting. Note this is just the initial implementation.

## Changes
- Firebase Analytics setup for iOS and Android
- Firebase Crashlytics setup for iOS and Android for uncaught, Flutter and Crashlytics trapped errors
- Analytics flagging each screen in the app as a screen, using constants and separate to navigation observer to ensure maintainable data in analytics as the app changes.

## Screenshots
No UI changes.

## Things to note
- Release notes not updated as no user facing changes.
- Analytics event method is present but not yet applied to event like playing PPE videos
- Firebase console and Google Analytics console setup also accompany this PR.
- Removed an orphan/duplicate file


lib/constants.dart
```diff
@@ -5,4 +5,36 @@ class Constants {
   static const String settingDisclaimerVersion = 'disclaimer_version';
   static const String settingDisclaimerAgreedDateTime =
       'disclaimer_agreed_date_time';
+
+  /// Analytics screen names avoid changing to keep analytics data constant
+  /// Even if the UI screen name changes try to keep this as analytics name
+  static const String analyticsHomeScreen = 'home';
```
maks, line 11: can these just be the route name consts to keep things more DRY ?

lib/utils/firebase.dart
```diff
@@ -0,0 +1,83 @@
...
+    // This is only to be used for confirming that reports are being
+    // submitted as expected. It is not intended to be used for everyday
+    // development.
+    Crashlytics.instance.enableInDevMode = false;
+
+    // Pass all uncaught errors from the Flutter framework to Crashlytics
+    FlutterError.onError = Crashlytics.instance.recordFlutterError;
+  }
+
+  /// WARNING - Leave these here and commented out unless simulating errors
```
maks: could we at least just move this into a CrashSimlutator class please? if we want to keep code then lets make sure its not commented out so all the usual tooling works with it.

lib/utils/firebase.dart
```diff
@@ -0,0 +1,83 @@
...
+  //    Future<void>.delayed(
+  //      const Duration(seconds: 2),
+  //      () {
+  //        List<int> list = <int>[];
+  //        print(list[100]);
+  //      },
+  //    );
+  //  }
+
+  CrashReporting() {
```
maks: Its a bit strange to have a separate init method thats only called from the constructor, is there a reason for that?

lib/view/airway/checklist/intubation_checklist_infographic_page.dart
```diff
@@ -1,11 +1,17 @@
...
 
+import '../../../constants.dart';
 import '../../../style.dart';
+import '../../../utils/firebase.dart';
 import '../../reusable_infographic_page.dart';
 
 class IntubationChecklistInfographicPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
+    // Analytics set screen name, stays until another screen changes it
```
maks, line 11: I'm not very familiar with Google analytics client, but if its already tracking route navigation events, what does this add as there are no other analytics events being tracked on this screen?

lib/view/airway/checklist/intubation_checklist_page.dart
```diff
@@ -25,6 +27,9 @@ class IntubationChecklistPage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    // Analytics set screen name, stays until another screen changes it
+    Analytics().analyticsScreen(Constants.analyticsIntubationChecklistScreen);
```
maks: ditto above for all the other calls to Analytics().analyticsScreen()

lib/main.dart
```diff
@@ -31,7 +35,17 @@ import 'view/staff_welfare/your_welfare_view.dart';
 import 'view/view_templates/html_text_card_view_template.dart';
 
 void main() {
-  runApp(const MyApp());
+  // Add this, and it should be the first line in main method to ensure no crashes before runApp()
+  // Currently Crashlytics and SharedPreferences packages below use platform channels so need this before runApp()
+  WidgetsFlutterBinding.ensureInitialized();
+
+  // Crash Reporting configured for not reporting whilst doing Dev mode builds
+  CrashReporting();
```
maks: it looks quite strange just calling a constructor and then not doing anything with the object itself. Since CrashReporting has no instance properties can it just be a static function?

ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -209,6 +210,24 @@
...
+			inputPaths = (
+				"$(BUILT_PRODUCTS_DIR)/$(INFOPLIST_PATH)",
+			);
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "# Type a script or drag a script file from your workspace to insert its path.\n${PODS_ROOT}/Fabric/run\n\n";
```
sallychung: Purely for project config cleanliness.
The comment in the script doesn’t make sense in here as it’s run in the build phase. Can we please remove it? 
Doesn’t have to be done in this PR

lib/view/icu_non_intensivist/daily_round/daily_round_view.dart
```diff
@@ -2,18 +2,22 @@ import 'package:flutter/material.dart';
...
 import '../../../models/icu_daily_round_checklist.dart';
+import '../../../strings.dart';
 import '../../../style.dart';
+import '../../../utils/firebase.dart';
 import '../../../widget/containers/icu_daily_round_container.dart';
 import '../../../widget/lists/checklist_card.dart';
 import '../../view_templates/tab_view_template.dart';
 
 class ICUDailyRoundView extends StatelessWidget {
-  final title = 'Daily Round';
```
maks, line 13: :+1:

maks (changes requested):

lib/constants.dart
workerbee22: @maks This was very much a conscious decision, because one of the poor people who have to review analytics, routes change a lot as the app changes, name changes, re routes, changed routes etc.Which means over a short time the route based method breaks down and it becomes a mess of new, newer, newest, older etc.

So this way the analytics screens should mostly stay fixed and independent as the routes all inevitably change.

lib/utils/firebase.dart
workerbee22: @maks So you'd prefer separation of initialisation and simulation?

lib/utils/firebase.dart
workerbee22: @maks Not really ... what would you prefer?

lib/view/airway/checklist/intubation_checklist_infographic_page.dart
workerbee22: See above ☝️

lib/view/airway/checklist/intubation_checklist_page.dart
workerbee22: See above ☝️

lib/main.dart
workerbee22: I'll check

lib/view/icu_non_intensivist/daily_round/daily_round_view.dart
workerbee22: Yep a reversion.

lukesleeman:
I've tested the branch out, and it appears to have no ill effects.  Kind of hard to tell if its doing anything or not 😁

workerbee22:
@lukesleeman Yep, until there is some data, it's a nice to have thing. 

But using the simulation code (in there but commented out) one crash type at a time and setting `Crashlytics.instance.enableInDevMode = true` you can see the crashes locally on your console. 

So I've tested all 3 levels of crash locally on Android and iOS for the simulations. Just waiting on crash data to show up in the console ... but its been a while so will go back and check. But if the app isn't talking to Crashlytics then when you goto the Firebase console you only get the setup wizard'. So because we can see the proper Crashlytics console view the app MUST have talked to Crashlytics already.

sallychung (approved):
Seems to be working. Tested through Proxyman

image

lib/constants.dart
maks: Hmm, ok, I've never been a consumer of analytics so will defer to those that do :+1:

lib/main.dart
maks: So consolidating the comment about the constructor with this one: can really just be a static function, no class needed

lib/utils/firebase.dart
maks: yes the simulation is only for local debugging and testing so if its useful to keep once this is ready for merging then yes please the simulation should be separate from the production code that does the initialisation

ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -209,6 +210,24 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		30531896244020B4005AE004 /* ShellScript */ = {
```
yimingc2, line 213: Is the script for Xcode 10? After 30th April, apps that submitted to App Store must use iOS 13 SDK or later which is Xcode 11+.

maks:
>So I've tested all 3 levels of crash locally on Android and iOS for the simulations. Just waiting on crash data to show up in the console ... but its been a while so will go back and check. But if the app isn't talking to Crashlytics then when you goto the Firebase console you only get the setup wizard'. So because we can see the proper Crashlytics console view the app MUST have talked to Crashlytics already.

@workerbee22  So I'd feel *much* better once we can actual app generated errors, especially app crashes in the console :slightly_smiling_face:  
as this is *exactly* what I experienced previously trying and failing to use Crashlytics in a Flutter app: everything looked fine locally but nothing ever showed up in the web console...

workerbee22:
I'll do a check against the project for which I provided the screen shots ie. these. Perhaps I missed something:

Issue: https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/15

image

yimingc2 (approved):

lib/utils/firebase.dart
```diff
@@ -0,0 +1,83 @@
...
+import 'package:firebase_crashlytics/firebase_crashlytics.dart';
+import 'package:flutter/foundation.dart';
+
+/// Analytics - Initialises and provides helper methods for Firebase Google Analytics
+class Analytics {
+  static FirebaseAnalytics analytics = FirebaseAnalytics();
+  static FirebaseAnalyticsObserver observer =
+      FirebaseAnalyticsObserver(analytics: analytics);
+
+  Future<void> analyticsScreen(String screenName) async {
```
yimingc2: We can make it `static` and then we don't need to create a new instance of this class every time.

yimingc2:
Looks good to me! Appreciate for your effort to make it happen!

ios/Runner.xcodeproj/project.pbxproj
workerbee22: Yep ... it was the default. I'll remove 👍

ios/Runner.xcodeproj/project.pbxproj
workerbee22: That was auto generated, so no idea.

lib/utils/firebase.dart
workerbee22: oh yeah should be ... oops 👍@maks flagged it too.

workerbee22:
OK so nothing in the console yet for Crashlytics from my tests ... investigating.

Analytics data coming through fine.

lib/utils/firebase.dart
workerbee22: Ok abstracted into `CrashSimulation` class.

lib/main.dart
workerbee22: ok just a static function now.

ios/Runner.xcodeproj/project.pbxproj
workerbee22: Removed comment.

lib/utils/firebase.dart
workerbee22: done

workerbee22:
OK PR workflow build all OK. Merging this one.

PR #239, opened 2020-04-10
[WIP] Potential fix for the app bar flashing

lukesleeman:
There appears to be this weird flashing of the app bar under iOS as the user scrolls down.

Under android the colour of the app bar as the user scrolls is darker than it should be.

I suspect both of these issues are caused by the elevation of the app bar.  It seems to change from 0 to fully elevated once its colour is not transparent.

This fix changes the elevation so it only switches to elevated once the app bar background colour is fully opaque.  🤞Hopefully this will resolve the flicker issue.

- [ ] `release-notes.txt` has been updated.

## Changes

- Only change elevation once the scroll transition is complete

## Screenshots

On android the colour of the app bar during the transition is not so dark 🎉:
image
## Things to note

I am not a flutter developer and I have no idea what I am doing 🤷‍♂️

### DIFF
--- a/lib/view/home_page.dart
+++ b/lib/view/home_page.dart
@@ -84,6 +84,7 @@ class _HomePageState extends State<HomePage> {
     final percentage = scrollPos < appBarBottom
         ? ((scrollPos) / appBarBottom.toDouble())
         : 1.00;
+    final elevation = percentage < 1.00 ? 0.00 : 4.00;
 
     return Builder(builder: (context) {
       final _scrollController = PrimaryScrollController.of(context);
@@ -127,6 +128,7 @@ class _HomePageState extends State<HomePage> {
             floating: true,
             pinned: true,
             snap: false,
+            elevation: elevation,
           ),
           _renderList(context)
         ],


lib/view/home_page.dart
```diff
@@ -84,6 +84,7 @@ class _HomePageState extends State<HomePage> {
     final percentage = scrollPos < appBarBottom
         ? ((scrollPos) / appBarBottom.toDouble())
         : 1.00;
+    final elevation = percentage < 1.00 ? 0.00 : 4.00;
```
lukesleeman, line 87: According to the docs
> If this property is null, then [ThemeData.appBarTheme.elevation] is used, if that is also null, the default value is 4, the appropriate elevation for app bars.

It would be nice to be able to reference `ThemeData.appBarTheme.elevation` rather than hard coding to 4, but I have no idea how to do that 🤷 

Perhaps one of the more experienced flutter devs could suggest a change

lukesleeman:
Ok, it seems that this doesn't fix the issue on iOS.  I've logged it under https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/242 for somebody with more flutter experience to look into

lib/view/home_page.dart
kimngyn: Would FlexibleSpaceBar work?

PR #244, opened 2020-04-11
[Fix] Rich Text Font Accessibilty

### LINKED ISSUE #227: Minor copy change and link insertion to Disclaimer @conditions of use
1. After "This app is for clinicians and hospital staff use ONLY"
Change the first mention of "WHAC19" to "Western Health Anaesthesia COVID-19 (WHAC19)" 

Thereafter it can be referred to WHAC19. This just ensures users are familiar with the origins of the acronym.

2. Insertion of a new sentence below section:

" WHAC19 aims to provide a really quick, usable means to access the core information for this."

Insert:

"Please refer to https://coronavirus.wh.org.au/ to ensure you have the latest up-to-date information on Western Health's Coronavirus guidelines."

This can be a standalone sentence above "Please keep in mind".

@Western-Health-Covid19-Collaboration/flutter-devs

yimingc2:
Resolved #227

## Abstract
This PR will fix the issue that the paragraph with hyperlink on Disclaimer Screen does not change the font size based on the accessibility setting. RichText is a low-level way to draw text and does not support font size accessibility. RichText should not be used in general.

## Changes

- Replace RichText with Text.rich

## Screenshots
**_Note that in the screenshot, the font size in accessibility is set to the smallest._**

Simulator Screen Shot - iPhone 11 Pro - 2020-04-11 at 21 01 59

## Things to note
I haven't updated the disclaimer version because there's no text changes. It's just the style. To test it, the QA needs to delete the existing app and install it again to see the disclaimer page.

### DIFF
--- a/lib/view/disclaimer_view.dart
+++ b/lib/view/disclaimer_view.dart
@@ -108,8 +108,7 @@ class _DisclaimerViewState extends State<DisclaimerView> {
                           ' for this.',
                           style: Styles.textH5,
                         ),
-                        RichText(
-                            text: TextSpan(children: [
+                        Text.rich(TextSpan(children: [
                           const TextSpan(
                             text: '\n\nPlease refer to ',
                             style: Styles.textH5,


ericjiang97 (approved):

workerbee22:
I'm afraid I don't agree that **RichText** is a '...a low-level way to draw text ...' as it is Flutters only way to do mixed styling atm. Nowhere in the doco does it say that ... or that it should not be used generally that I have ever found. It is a standard widget like all the others.

workerbee22 (changes requested):
I don't think we need the TextSpan.rich constructor, as it just allows you to style multiple spans with the default text style (which we are not using here?) while still allowing specified styles per span (which we ar explicitly using here).

workerbee22:
We'll need to discuss the bigger issue of Accessibility, which we know we should all consider and treat, but the work is always significant to do this. So not sure its a priority right now?

Very much a summary only, but here it the 'tip of the iceberg' on Flutter and accessibility:

https://flutter.dev/docs/development/accessibility-and-localization/accessibility

yimingc2:
> I'm afraid I don't agree that **RichText** is a '...a low-level way to draw text ...' as it is Flutters only way to do mixed styling atm. Nowhere in the doco does it say that ... or that it should not be used generally that I have ever found. It is a standard widget like all the others.

I read about the `RichText` and `Text.rich` from here https://api.flutter.dev/flutter/widgets/Text/Text.rich.html

The issue of `RichText` is that it does not support font size accessibility while `Text.rich` does.

yimingc2:
> I don't think we need the TextSpan.rich constructor, as it just allows you to style multiple spans with the default text style (which we are not using here?) while still allowing specified styles per span (which we ar explicitly using here).

I'm using different TextStyle for the blue hyperlink in the third paragraph. Yeah I think we can discuss about accessibility later.

workerbee22:
> > I'm afraid I don't agree that **RichText** is a '...a low-level way to draw text ...' as it is Flutters only way to do mixed styling atm. Nowhere in the doco does it say that ... or that it should not be used generally that I have ever found. It is a standard widget like all the others.
> 
> I read about the `RichText` and `Text.rich` from here https://api.flutter.dev/flutter/widgets/Text/Text.rich.html
> 
> The issue of `RichText` is that it does not support font size accessibility while `Text.rich` does.

Ok I'll take your word for it as I assume you've tested it. Also keen if you have a Flutter issue for this as I have another app this would impact 👍

workerbee22 (approved):

PR #247, opened 2020-04-12
238 - Move UI config out of data layer

### LINKED ISSUE #238: UI classes need to be removed from data layer
For now `hard_data.dart` and the classes within it are our data layer so UI classes like `ReusableCard` have no business being there and needed to be refactored into separate datamodels holding the data, which is displayed by the `ReusabelCard` widgets

maks:
Refactor that removes all UI configuration from the data layer (which is currently just hardcoded in `hard_data.dart`)

## Screenshots

There should be no visual changes to the existing app as the refactor just cleans up the app code architecture.

Fixes: #238

### DIFF
--- a/lib/hard_data.dart
+++ b/lib/hard_data.dart
@@ -11,7 +11,6 @@ import 'models/ventilation_guide.dart';
 import 'routes.dart';
 import 'strings.dart';
 import 'style.dart';
-import 'widget/reusable_card.dart';
 
 /// Web urls
 const String whURL = 'http://wh.cyphix.net/';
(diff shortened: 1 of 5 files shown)


workerbee22 (approved):
LGTM

lib/hard_data.dart
```diff
@@ -1,17 +1,15 @@
-// A list of two cards
-import 'package:flutter/material.dart';
+// import 'package:flutter/material.dart';
```
domesticmouse, line 1: nit: prefer deleting outright to commenting out. We use source control for a reason =)

domesticmouse (commented):
LGTM

lib/hard_data.dart
maks: Doh! thanks for spotting that @domesticmouse , yes that's a plain mistake I'm the same - no leaving commented out code!
Unfort this has been merged, I'll make sure to fix that in my next PR !

Issue #196, opened 2020-04-06
iOS Version - App Store Connect Reported Issues


workerbee22:
When we submit to App Store Connect (used for TestFlight and the App Store) Apple analyses the app files and reports on any issues.

As at v1.0.101 we have these issues. Currently neither of these is stopping the build nor the release of the app for testing on iOS:

![image](https://user-images.githubusercontent.com/5154664/78614613-555ea800-78b2-11ea-8f7c-17d0f7568df5.png)

workerbee22:
**ITMS-90078: Missing Push Notification Entitlement**

Is a Flutter framework issue, that is being worked on but the consensus is to ignore it for now. See Flutter issue:  https://github.com/flutter/flutter/issues/9984

workerbee22:
**ITMS-90381: Too many symbol files**

Crept in as a new issue with v1.0.101 when I locally updated brew, ruby and cocoapods to make sure we were using cocoapods v1.9.1. 

There are fixes for this that I'm investigating such as this and others: https://stackoverflow.com/questions/34313049/too-many-symbol-files-warning-when-submitting-app

This may be caused by our change in v1.0.101 to only support iOS 12.0 and above.

yimingc2:
> **ITMS-90381: Too many symbol files**
> 
> Crept in as a new issue with v1.0.101 when I locally updated brew, ruby and cocoapods to make sure we were using cocoapods v1.9.1.
> 
> There are fixes for this that I'm investigating such as this and others: https://stackoverflow.com/questions/34313049/too-many-symbol-files-warning-when-submitting-app
> 
> This may be caused by our change in v1.0.101 to only support iOS 12.0 and above.

@workerbee22 , the reason of the warning is that iOS deployment target for the Pods is iOS8 which is lower than the project deployment target iOS12. So it would generate armv7 dSYM files but no binaries.
So the solution is add 
`config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '12.0'`
to the Podfile.

workerbee22:
Awesome thanks @yimingc2 . I'm also looking at this SO article to confirm syntax etc: https://stackoverflow.com/questions/54704207/the-ios-simulator-deployment-targets-is-set-to-7-0-but-the-range-of-supported-d

So this is as per your solution above June and seems to be the best answer:

[...]

PR #249, opened 2020-04-13
Fix for Too many symbols Apple warning

workerbee22:
## Abstract

Fix for part of the issue [196](https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/196) where App Store Connect would give a warning about too many symbols:

image

## Changes

- Podfile fix to make sure all packages adhere to iOS 12.0 minimum version.

## Screenshots

No UI changes

## Things to note

- Retested on iOS device to ensure it builds, runs and causes no new issues.
- Not linking to issue 96 as there are other warnings still to be fixed in this issue.


yimingc2 (approved):
LGTM! Archived from Xcode and checked all the generated dSYM files for third party libs and it looks good. Good job!!

workerbee22:
> LGTM! Archived from Xcode and checked all the generated dSYM files for third party libs and it looks good. Good job!!

Thanks for looking into this 👍

PR #265, opened 2020-04-16
Implement Acknowledgements View

### LINKED ISSUE #148: Build out the Acknowledgements and references screen
We need to build out an Acknowledgements and references screen

rockgecko-development:
Implement #148 Acknowledgements View

## Screenshots

image

image

### DIFF
--- a/lib/view/acknowledgements_view.dart
+++ b/lib/view/acknowledgements_view.dart
@@ -1,0 +1,118 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_widget_from_html/flutter_widget_from_html.dart';
+import '../constants.dart';
+import '../models/home_card.dart';
+import '../strings.dart';
+import '../style.dart';
+import '../utils/firebase.dart';
+import '../widget/cards/reusable_card_base.dart';
+import '../widget/reusable_card.dart';
+
+class AcknowledgementsView extends StatelessWidget {
+  static const _cards = [
+    'References',
+    HomeCard(
+        title: 'Safe Airway Society',
+        description: 'Guidelines and Infographics',
+        route: 'https://www.safeairwaysociety.org/covid19/'),
+    HomeCard(
+        title: 'Intensive Care Society',
+        description: 'Wellbeing Guidelines',
+        route:
+            'https://www.ics.ac.uk/ICS/Education/Wellbeing/ICS/Wellbeing.aspx'),
+    HomeCard(
+        title: 'Australian New Zealand Intensive Care Society',
+        description: 'COVID-19 Guidelines',
+        route: 'https://www.anzics.com.au/coronavirus-guidelines/'),
+    HomeCard(
+        title: 'Western Health Department of Anaesthesia',
+        description: 'Policies and Procedures',
+        route: null),
+    HomeCard(
+        title: 'Western Health Novel Coronavirus Information',
+        description: 'COVID-19 Information',
+        route: 'https://coronavirus.wh.org.au/'),
+    HomeCard(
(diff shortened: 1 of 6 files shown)


workerbee22:
I'd love to improve this via:
- use all the organisational logos here 
- break into sections a little more
- also prefer RichText() widgets here rather than html given we are just adding paragraph breaks and tappable links, like June did for the Disclaimer link.

But as an initial implementation it totally works 👍

lib/view/acknowledgements_view.dart
```diff
@@ -0,0 +1,118 @@
...
+import '../widget/cards/reusable_card_base.dart';
+import '../widget/reusable_card.dart';
+
+class AcknowledgementsView extends StatelessWidget {
+  static const _cards = [
+    'References',
+    HomeCard(
+        title: 'Safe Airway Society',
+        description: 'Guidelines and Infographics',
+        route: 'https://www.safeairwaysociety.org/covid19/'),
```
workerbee22, line 18: Perhaps we should use Strings values for the URLs? Because these are likely to change often.

workerbee22: ditto for al the urls

lib/view/acknowledgements_view.dart
```diff
@@ -0,0 +1,118 @@
...
+                    child: ReusableCardBase(
+                      child: [
+                        HtmlWidget(
+                          _thanksHtml,
+                          textStyle: Styles.textP,
+                        )
+                      ],
+                    ),
+                  ),
+                ]))));
```
workerbee22, line 117: Trailing commas please, it makes my IDE much happier with the Flutter plugin 😁

lib/view/acknowledgements_view.dart
```diff
@@ -0,0 +1,118 @@
...
+                        e.toString(),
+                        style: Styles.textH4,
+                      ),
+                    );
+                  }),
+                  Padding(
+                    padding: const EdgeInsets.only(top: 4),
+                    child: ReusableCardBase(
+                      child: [
+                        HtmlWidget(
```
workerbee22, line 109: Ideally would prefer RichText to get styled links and paragraph breaks. But I think this is fine for now 👍

workerbee22 (changes requested):
@rockgecko-development Good and fast job !  Just some minor suggestions.

lib/view/acknowledgements_view.dart
kimngyn: Agreed with @workerbee22 regarding trailing commas ☝️ 
Everything else LGTM 😄

lib/view/acknowledgements_view.dart
rockgecko-development: how would you do that for the `<a href` links in the bottom card? Don't think it's worth it, much easier for the card's link to be next to the card's title and description. We don't refer to these urls anywhere else so no reuse benefit.

lib/view/acknowledgements_view.dart
rockgecko-development: You can style the links with `hyperlinkColor`. Looks like in the mock it's a slightly lighter blue.
Paragraph breaks match the design as far as I see, what are you referring to?

lib/view/acknowledgements_view.dart
workerbee22: True no re-use.

lib/view/acknowledgements_view.dart
workerbee22: Leave it for now because we use this approach everywhere.  This is bigger discussion of RichText vs html/markdown 🤣

workerbee22 (approved):

PR #266, opened 2020-04-16
add sentry error reporting

maks:
Add Sentry reporting for app across Flutter, native Android and iOS.

Includes refactor, renaming the App widget and moving it into its own file.

This has no impact on user visible UI or functionality.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -74,4 +74,8 @@ flutter {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
+    implementation 'io.sentry:sentry-android:1.7.27'
+    implementation 'org.slf4j:slf4j-nop:1.7.25'
+    
 }
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -6,7 +6,7 @@
          additional functionality it is fine to subclass or reimplement
          FlutterApplication and put your custom class here. -->
     <application
-        android:name="io.flutter.app.FlutterApplication"
+        android:name=".WHApplication"
         android:label="WHAC19"
         android:icon="@mipmap/ic_launcher">
         <activity
--- a/android/app/src/main/kotlin/org/whegba/wh_covid19/WHApplication.kt
+++ b/android/app/src/main/kotlin/org/whegba/wh_covid19/WHApplication.kt
@@ -1,0 +1,13 @@
+package org.whegba.wh_covid19
+
+import io.flutter.app.FlutterApplication
+import io.sentry.Sentry
+import io.sentry.android.AndroidSentryClientFactory
+
+class WHApplication : FlutterApplication() {
+
+    override fun onCreate() {
+        super.onCreate()
+        Sentry.init(AndroidSentryClientFactory(this))
+    }
+}
--- a/ci_cd.md
+++ b/ci_cd.md
@@ -1,0 +1,12 @@
+### Sentry config
+
+The DSN needs to be supplied by the build environment.
+
+For native Android, this is via a properties file in: `app/src/main/resources` which is created by the following custom script in Codemagic and for Flutter, this is via a Dart file created at build time also by this script:
+```shell
+#!/usr/bin/env sh
+set -e # exit on first failed commandset
+mkdir -p $FCI_BUILD_DIR/android/app/src/main/resources
+echo dsn=$SENTRY_DSN > $FCI_BUILD_DIR/android/app/src/main/resources/sentry.properties
+echo "const DSN='$SENTRY_DSN';" > $FCI_BUILD_DIR/lib/env.dart
+```
--- a/lib/error_reporting.dart
+++ b/lib/error_reporting.dart
@@ -14,7 +14,7 @@ Future<SentryClient> get _sentryClient async => SentryClient(
 
 Future<Event> get _environmentEvent async {
   final packageInfo = await PackageInfo.fromPlatform();
-  final deviceInfoPlugin = await DeviceInfoPlugin();
+  final deviceInfoPlugin = DeviceInfoPlugin();
   OperatingSystem os;
   Device device;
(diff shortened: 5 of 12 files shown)


android/app/build.gradle
```diff
@@ -74,4 +74,8 @@ flutter {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
+    implementation 'io.sentry:sentry-android:1.7.27'
```
workerbee22, line 78: So is this required to catch the Android platform level stuff? This seems to be an Android Sentry implementation?

android/app/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,7 @@
          additional functionality it is fine to subclass or reimplement
          FlutterApplication and put your custom class here. -->
     <application
-        android:name="io.flutter.app.FlutterApplication"
+        android:name=".WHApplication"
```
workerbee22, line 9: Do we have a reason for changing this?

android/app/src/main/kotlin/org/whegba/wh_covid19/WHApplication.kt
```diff
@@ -0,0 +1,13 @@
+package org.whegba.wh_covid19
```
workerbee22, line 1: Kotlin code? Again I assume this is to implement the Android platform side of it?

ci_cd.md
```diff
@@ -0,0 +1,12 @@
+### Sentry config
```
workerbee22, line 1: Assume this goes along with codemagic changes? Which script in codemagic? Make sure you don't clobber the Post script we have atm for the Firebase App Dist things.

lib/error_reporting.dart
```diff
@@ -0,0 +1,78 @@
...
+import 'package:device_info/device_info.dart';
+import 'package:flutter/foundation.dart';
+import 'package:package_info/package_info.dart';
+import 'package:sentry/sentry.dart';
+import 'env.dart';
+
+Future<SentryClient> get _sentryClient async => SentryClient(
+      dsn: dsn,
+      environmentAttributes: await _environmentEvent,
+    );
```
workerbee22, line 13: dartfmt formatted?

lib/error_reporting.dart
```diff
@@ -0,0 +1,78 @@
...
+      version: androidInfo.version.release,
+    );
+    device = Device(
+      model: androidInfo.model,
+      manufacturer: androidInfo.manufacturer,
+      modelId: androidInfo.product,
+    );
+  } else if (Platform.isIOS) {
+    final iosInfo = await deviceInfoPlugin.iosInfo;
+    os = OperatingSystem(
```
workerbee22, line 33: Good these are the best values for iOS 👍

lib/error_reporting.dart
```diff
@@ -0,0 +1,78 @@
...
+      name: iosInfo.systemName,
+      version: iosInfo.systemVersion,
+    );
+    device = Device(
```
workerbee22, line 37: ditto 👍

lib/error_reporting.dart
```diff
@@ -0,0 +1,78 @@
...
+        name: packageInfo.appName,
+        version: packageInfo.version,
+        build: packageInfo.buildNumber,
+      ),
+    ),
+  );
+  return environment;
+}
+
+void flutterOnErrorHandler(FlutterErrorDetails details) {
```
workerbee22, line 59: Cool uncaught errors 👍

lib/main.dart
```diff
@@ -1,137 +1,37 @@
...
-import 'view/ppe/ppe_view.dart';
-import 'view/sbs_guide_view.dart';
-import 'view/staff_welfare/your_welfare_view.dart';
+import 'error_reporting.dart';
+import 'wh_app.dart';
 
 void main() {
   // Add this, and it should be the first line in main method to ensure no crashes before runApp()
-  WidgetsFlutterBinding.ensureInitialized();
+  //WidgetsFlutterBinding.ensureInitialized();
```
workerbee22: Do we need to comment this out? Does it cause issues? I've found it helps prevent ugly issues at app startup.

lib/main.dart
```diff
@@ -1,137 +1,37 @@
...
 
-  runApp(const MyApp());
-}
+  // This captures errors reported by the Flutter framework.
+  FlutterError.onError = flutterOnErrorHandler;
+
+  // Lets not show end users the red screen of death
```
workerbee22, line 16: Yep ...

workerbee22 (approved):
LGTM if a bit more complicated than I'm used to for Sentry.

workerbee22:
@maks There is no more iOS setup I can do quickly to trap iOS platform errors, like you have here for Android. So no more commits from me on this for iOS right now.

android/app/build.gradle
maks: Yes we want to be able to catch errors that happen in native Android code, either in Flutter framework or plugins, which won't necessarily be reported to Dart.

android/app/src/main/AndroidManifest.xml
maks: Yes, we need a custom Android App singleton class in order to do native android Sentry initialisation in its `onCreate()`

android/app/src/main/kotlin/org/whegba/wh_covid19/WHApplication.kt
maks: yep see my comment above about native Android sentry init

ci_cd.md
maks: Ah yes I should include here that it needs to be a `pre-build` step Codemagic workflow script

lib/main.dart
maks: Oops, I thats a mistake, I meant to put that back.

android/app/src/main/AndroidManifest.xml
workerbee22: Yeah cool 👍

PR #272, opened 2020-04-18
Fixed the Home page 'flickering' of the app bar

### LINKED ISSUE #242: Fix flicker with app bar on iOS
On iOS the app bar flickers in and out as you scroll up and down

**iOS app bar flicker**
Screen-Recording-2020-04-10-at-5 43 12-pm

Meanwhile on android the situation is a little bit better, however you still get an un-naturally dark background behind the app bar as you scroll:

**Android app bar black background**

android app bar 2

I _strongly suspect_ this all has something to do with the elevation of the app bar.  It seems to automatically set its self to have a elevation of 0 when it has a transparent background, then once the colour changes as we scroll it jumps to having an elevation of 4dp

I tried to create a PR which fixes the issue:
https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/pull/239

However it doesn't seem to solve the problem on iOS.  I'm guessing we will need a more experienced flutter developer to look at the issue.

workerbee22:
Fix for issue [242](https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/242)

## Abstract

- Fixed the 'flickering' of the app bar when scrolling on iOS
- Was due to over scroll on iOS permitting the scroll position to be negative

## Changes

- general cleanup of the SliverAppBar
- ensure that scroll position cannot be negative, because on iOS when overscrolling the value can be negative
- elevation on SliverAppBar and other cleanups

## Screenshots

See issue for the problem

## Things to note

Retested on Android and iOS.

### DIFF
--- a/lib/view/home_page.dart
+++ b/lib/view/home_page.dart
@@ -108,66 +108,78 @@ class _HomePageState extends State<HomePage> {
   }
 
   Widget _renderBody(BuildContext context) {
-    const appBarHeight = 70;
-    const logoHeight = 42;
+    const appBarHeight = 70.0;
+    const logoHeight = 42.0;
 
-    final Widget mainLogo = SvgPicture.asset('assets/images/main_logo.svg',
-        height: logoHeight.toDouble());
-    final appBarBottom = appBarHeight.toDouble() - logoHeight.toDouble();
-    final scrollPos = _scrollPosition.round();
-    final percentage = scrollPos < appBarBottom
-        ? ((scrollPos) / appBarBottom.toDouble())
-        : 1.00;
+    final Widget mainLogo = SvgPicture.asset(
+      'assets/images/main_logo.svg',
+      height: logoHeight,
+    );
+    const appBarBottom = appBarHeight - logoHeight;
+    // On iOS scroll position can be negative due to over scroll
+    // So prevent negative scrollPos which results in a negative percentage
+    final scrollPos = _scrollPosition.isNegative ? 0.0 : _scrollPosition;
+    final percentage =
+        scrollPos < appBarBottom ? scrollPos / appBarBottom : 1.0;
 
-    return Builder(builder: (context) {
-      final _scrollController = PrimaryScrollController.of(context);
+    return Builder(
+      builder: (context) {
+        final _scrollController = PrimaryScrollController.of(context);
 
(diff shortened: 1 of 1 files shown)


lib/view/home_page.dart
```diff
@@ -108,66 +108,78 @@ class _HomePageState extends State<HomePage> {
...
+            SliverAppBar(
+              // Warning brightness interacts with SystemUiOverlayStyle
+              // See system_bars.dart comments
+              brightness: Brightness.light,
+              primary: true,
+              automaticallyImplyLeading: false,
+              expandedHeight: appBarHeight,
+              centerTitle: false,
+              forceElevated: false,
+              elevation: 0.0,
```
lukesleeman, line 146: This means that at least on android, we loose the shadow behind the app bar:

image

However this does seem to be an acceptable tradeoff for solving the problem with black shadow on android.

You might want to have a look at this PR https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/pull/239
for a slightly better solution - at least with this one we have the elevation once we have scrolled along

lukesleeman (approved):
I've got one small comment about the elevation shadow on the app bar, but its a pretty minor thing.  Up to you if you want to play with it a bit more, or just merge 😁

Also, great to see this issue _finally_ solved 🙌

lib/view/home_page.dart
lukesleeman: Actually … I think there may be been some problems with the PR I’ve linked, in that scrolling fast seemed to stuff it up.  If you do go down this route be sure to give it some good testing, so make sure my approach doesn't just stuff things back up

workerbee22: Yep I changes elevation to 0.0 as it is also 0.0 on the nav bar. But I can put it back if you prefer.

workerbee22: I can remove elevation easily and put it back. Doing it now.

The fix wasn't actually anything to do with elevation.

workerbee22: forcedElevation doesn't actually do anything in this context so I'll remove it as well.

workerbee22: OK so still some things to fix later when elevation is present. But for now eleavtion is 0.0.

PR #275, opened 2020-04-18
[Enhancement] PPE UI Polish

### LINKED ISSUE #206: PPE - UI Inconsistencies
- [x] A Donning & Doffing Step by Step - Bullet points text colour should be #blackAlpha900
- [x] B Doffing - Both methods - Warning - Background color in warning should be 236 243 240, text color #blackAlpha900
- [x] C Doffing - Both methods - Warning - Margin between warning and Step 1 is too large
- [ ] D Doffing Method One - Step 5 and Step 3 reversed?
- [x] E Toolbar colours don't match zeplin - donning should be green, doffing should be purple

image

**Toolbars from zeplin**
image
image

yimingc2:
Resolved #206

## Abstract
This PR will update the color and style for the PPE page.

## Changes

- Remove unnecessary padding
- Add new text style
- Change the navigation bar color for Method 1 and 2
- Change the background color for Method 1 and 2

## Things to note
I don't think it's a good idea to use "new line" after multiline string to add bottom padding.

### DIFF
--- a/lib/hard_data.dart
+++ b/lib/hard_data.dart
@@ -86,8 +86,7 @@ class HtmlTextScreenData {
 /// PPE - all the PPE content
 const String ppeOffWarning = '''
 Outside of PPE is contaminated!
-If your hands get contaminated during PPE removal, immediately wash your hands or use an alcohol-based hand sanitizer.
-''';
+If your hands get contaminated during PPE removal, immediately wash your hands or use an alcohol-based hand sanitizer.''';
 
 const List<PPEStepInfo> ppeOnSteps = [
   PPEStepInfo(
--- a/lib/style.dart
+++ b/lib/style.dart
@@ -135,6 +135,13 @@ abstract class Styles {
     fontFamily: fontFamily,
   );
 
+  static const TextStyle textBody = TextStyle(
+    color: AppColors.majorText,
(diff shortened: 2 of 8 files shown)


lukesleeman:
👋 Hey @yimingc2 - great work on this!  I've run the app in an emulator and it seems like there is only one small thing missing:  In the donning step by step guide, the app bar is still grey, but it should be green.

**Current app**
image

**Zeplin**
image

lukesleeman:
Oh!  I also just realised that in zeplin we have an X button, but in the app we have a back button!  Better change it to an X!

yimingc2:
> Oh! I also just realised that in zeplin we have an X button, but in the app we have a back button! Better change it to an X!

Good catch Luke! I reckon whether to use X or Back button depends on how we show the screen. If we navigate to it (the screen comes from right to left),  back button is a good choice. If we present it (the screen comes from bottom to top), X button fits more. An example of presenting a screen is when the user press "PDF" button. It's just my point of view though.

lukesleeman (approved):
I've re-tested and its all looking good ✅

ericjiang97 (approved):

workerbee22 (approved):
LGTM

Issue #305, opened 2020-04-30
Button Behaviour Inconsistencies


workerbee22:
Behaviour on tap varies and is not consistent across the app. Key example are home screen shere call icon shows a ripple on pressed, but information has no on pressed feedback at all.

![image](https://user-images.githubusercontent.com/5154664/80670221-a9436200-8ae9-11ea-9750-bd8f204e3c4f.png)

lukesleeman:
I checked out all our toolbar buttons and its only the info icon which doesn't have the ripple - everywhere else seems to have it properly.

Since it seemed simple, I produced a small fix!

lukesleeman:
Based off what @workerbee22 has said, we can close this one off now

sallychung:
✅ v1.0.168

Seeing the ripple effect on both buttons

PR #308, opened 2020-04-30
Fix for info icon not showing ripple

lukesleeman:
This fixes issue #305 - I have no idea what I'm doing with flutter but I had a look at how the call button was done, and copied that!

Hopefully this doesn't break anything 🤷‍♂️

## Changes

- Replace gesture detector with IconButton to gain ripple effect

## Screenshots

ripple
## Things to note

I don't know what I'm doing 🙃


workerbee22:
Actually @lukesleeman there is a better fix. 

But on closer look it seems fine, but might optimise a little ... looking closer now.

workerbee22:
@lukesleeman Your change was fine 👍 but I did a little more cleanup of the code. Would like to do more but it has wider implications and so should wait for bigger changes post version 1.

workerbee22 (approved):
LGTM

PR #312, opened 2020-05-02
[Feature] Settings support and Privacy Setting

### LINKED ISSUE #301: Build settings screen which explains what data we gather and allows opt out
As part of our push to Make app GDPR compliant + better inform our users about what data is gathered  ( #241 ) we need to build a settings screen, under the info icon, that clearly explains what data we gather, what its used for and provides a switch to opt out.   The opt out switch will disables analytics and Sentry.

In terms of what we gather and why:

- Analytics on how the users are using the app with, gathered with google analytics
- Crash reports gathered with Sentry
- We gather this information to help is improve the app.  We hope to understanding what our users are using the app for, as well as proactively address any crashes that come up.

When users opt out, we can disable analytics by calling [setAnalyticsCollectionEnabled](https://pub.dev/documentation/firebase_analytics/latest/firebase_analytics/FirebaseAnalytics/setAnalyticsCollectionEnabled.html) 

With sentry @workerbee22 has some ideas on how we can disable crash reports

workerbee22:
Feature to add Settings for the app, allowing use selectable settings to be persisted to device storage and used on subsequent app startups.

Initially has just the Setting for Privacy. This setting allows the user to switch off data collection for analytics and crash reporting data.

## Changes

- Added tweak for trailing commas.
- Fixed 'acknowledgements' spelling mistake
- `release-notes.txt` has been updated.

## Architecture Change

- Read Settings values at app startup and set app state for these values (app state continue to use Provider package)
- These app state values are used multiple places through the code for each setting
- Setting switch changes the app state and also persists the new setting value to storage

The above approach is used rather than read the value from persistent storage at multiple places throughout the code. This minimises the reads/writes from/to persistent storage. This is preferred as many devices have slow storage speeds.

Settings values are persisted to storage as two separate values for analytics and crash reporting collection.  This is aggregated into one setting for Privacy in the UI. This allows for possible future separation ie. disable just analytics reporting but not crash reporting. This is important future proofing, because if persist just one setting for both, will be much harder to separate later, as this is stored on ALL existing installed devices.

## Screenshots

- **Privacy** card in **Information** screen moved to the the top, as probably more important to users than the others. Keep it named as **Privacy** for now, as it is the only setting. If other settings get added will change to **Settings**.

image

- Slightly tweaked text from design; changed 'and' to '&' and added para break and reduced words of last sentence.

image

## Things to note

- Changing app startup behaviour to allow for more than just one Setting
- Privacy is the only setting at the moment, but code to allow for more Settings in the future


workerbee22:
Just getting going on this now, will complete today via multiple commits.

lib/utils/storage.dart
```diff
@@ -0,0 +1,23 @@
...
+/// Persistent Storage on device read/write
+/// Read Settings values at app startup and set app state for these values
+/// (app state continue to use Provider package), These app state values are
+/// used multiple places through the code for each setting. This is approach
+/// is preferred rather than read the value from persistent storage at multiple
+/// places throughout the code. This minimises the reads/writes from/to
+/// persistent storage as devices have slow storage speeds.
+
+class Settings {
+  static Future<bool> readDisclaimerAgreed() async {
```
maks: making these static is going to make it impossible to test code that uses the settings. It would be better to make these class methods and then just have a Provider at the top of the widget tree to give access to whatever parts of the app need it.

lib/view/disclaimer_view.dart
```diff
@@ -100,24 +100,29 @@ class _DisclaimerViewState extends State<DisclaimerView> {
...
-                                ..onTap = () {
-                                  launch(whCoronavirusInfoURL);
-                                }),
-                          const TextSpan(
-                            text:
-                                ' to ensure you have the latest up-to-date information on Western Health\'s Coronavirus guidelines.\n\n',
-                            style: Styles.textH5,
+                            children: [
+                              const TextSpan(
+                                text: '\n\nPlease refer to ',
```
maks: better to use padding/margin vs embedded newlines. also shouldn't this be in string consts?

lib/view/disclaimer_view.dart
```diff
@@ -100,24 +100,29 @@ class _DisclaimerViewState extends State<DisclaimerView> {
...
+                                text: whCoronavirusInfoURL,
+                                style: Styles.textH5Hyperlink,
+                                recognizer: TapGestureRecognizer()
+                                  ..onTap = () {
+                                    launch(whCoronavirusInfoURL);
+                                  },
+                              ),
+                              const TextSpan(
+                                text:
+                                    ' to ensure you have the latest up-to-date information on Western Health\'s Coronavirus guidelines.\n\n',
```
maks, line 104: also strings const?

lib/widget/cards/reusable_card_base.dart
```diff
@@ -53,7 +54,7 @@ class ReusableCardBase extends StatelessWidget {
     final col = Column(
       mainAxisAlignment: verticalAlignment,
       crossAxisAlignment: CrossAxisAlignment.start,
-      mainAxisSize: MainAxisSize.min,
+      //mainAxisSize: MainAxisSize.min,
```
maks: needs to be deleted?

lib/widget/switch_card.dart
```diff
@@ -0,0 +1,113 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:url_launcher/url_launcher.dart';
+
+import '../hard_data.dart';
+import '../strings.dart';
+import '../style.dart';
+
+class SwitchCard extends StatefulWidget {
```
maks: I'm assuming this is intended to get wired up to the settings? If so once you do that if you expose the Settings via a Provider, this can become a stateless widget and the `_selected` state bool won't be needed.

maks (changes requested):
Great work @workerbee22 👍  I know you've marked this WIP, but just a bit of feedback on making sure the static stuff is removed.

lib/utils/storage.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import '../constants.dart';
+import '../strings.dart';
+
+/// Persistent Storage on device read/write
+/// Read Settings values at app startup and set app state for these values
+/// (app state continue to use Provider package), These app state values are
+/// used multiple places through the code for each setting. This is approach
+/// is preferred rather than read the value from persistent storage at multiple
+/// places throughout the code. This minimises the reads/writes from/to
```
maks: This isn't actually correct: on Android the SharedPrefs framework code already does in-memory read and write caching so no need attempt to replicate that in Flutter side. I think iOS equivalent may also do this.

maks (changes requested):

lib/utils/storage.dart
workerbee22: Yep planning that as per the PR Architecture changes comment. Still WIP.

lib/view/disclaimer_view.dart
workerbee22: Yeah sure.

lib/view/disclaimer_view.dart
workerbee22: Yeah ... but can say the same about ALL the hard_data.dart. So was going to wait until clarity around the Disclaimer, or just move to hard_data.

lib/widget/cards/reusable_card_base.dart
workerbee22: yep ... not needed.

lib/widget/switch_card.dart
workerbee22: Yep planning that as per the PR Architecture changes comment. Still WIP.

lib/utils/storage.dart
workerbee22: Ok didn't know that, but the point is still valid either way, should use Provider rather than lots of calls to SharedPreferences all over the code. Also as per comment, setting up for possible other future settings.

lib/view/disclaimer_view.dart
workerbee22: But not going to do that as part of this PR.

lib/view/disclaimer_view.dart
workerbee22: Not part of this PR, so so later in another PR when Disclaimer content issues are resolved.

lib/utils/storage.dart
workerbee22: @maks I though static methods were called 'class methods' ... so not exactly sure what you want here? And also already have a Provider at the top of the tree for app state. 

So the ways I've approached it is to have this Settings class that doe all the read/write to storage, but keep that separate from the app state models for provider. Keen to maintain separation of concerns.

lib/widget/switch_card.dart
workerbee22: Stateless widget now using Provider for app  state, but still have Settings as a separate class for read/write to storage.

lib/utils/storage.dart
maks: Oh complete agree on using Provider and abstracting shared prefs access 💯 on that. My feedback just on the read/write comment

lib/utils/storage.dart
maks: Mike the whole point of Provider is DI (dependency injection) so you don't use static, the fact that Provider helps with state mgmt is almost secondary concern. So for this case if you have static access to the settings class, how will you mock or stub it as a dependency in a test? If on # other hand you provide it via Provider, mocking in tests is straight forward.

workerbee22:
@maks and @rockgecko-development this is working fine but needs some cleanup and I need you guys to review and be happy with approach. I've changed my mind on this 3 times already so want your feedback as there are pros and cons for each approach. Specifically look at:

- `Settings` class is just for storage read/write for privacy setting and disclaimer details, and any future settings we have to add like for onboarding etc.
- Privacy setting switch uses provider, but no other UI changes besides the switch, so its more future proofing the approach, otherwise it's a bit of overkill.

@maks also need your thoughts on the best way to actually disable the crash reporting. At the moment just using the privacy setting directly from storage in the same place we check for release mode or not. But might be a better way.

Also analytics wise there are a few options for disabling with various pros/cons, so will get feedback on above before picking the method to disable.

lib/utils/storage.dart
workerbee22: Ok ... I'll have to investigate.

maks: @workerbee22 If it's easier, I can do implementation of this later tonight?

lukesleeman (approved):
I can't speak for the flutter side of things, but I have tested this out on android inside the emulator and it all seems ☑️

maks (approved):
I'm happy for this to land and then I'll do my architecture refactor PR afterwards

workerbee22:
Also fixed this issue with transitive dependency `extended_image` for the `native_pdf_view` package:

https://github.com/fluttercandies/extended_image/issues/160

workerbee22:
Still some cleanup and tests to add, but moving forward with this due to Beta 3 urgency.

workerbee22:
Completed this despite some cleanup and tests to add, to be done as separate PR. Other suggested implementation changes in a separate PR but changes no UI or result, so defer that until after Version 1 when other major re-architecting and improvements will be done.

PR #329, opened 2020-05-06
[Fix] Disclaimer Margin and Text Color

### LINKED ISSUE #323: Disclaimer margin and text colour
All text under this sections needs to be black. This is more obvious on iPad as the extra white space makes it can be harder to read for those that are visually challenged (me 🤓)

Margins also needs to be tightened up. I’ve highlighted one example showing headings and content having margins that much larger than what’s in Zeplin. (Note that Zeplin’s scale is different but the proportions still aren’t matching in the app)

Copy changes 012

yimingc2:
Resolved #323

## Abstract
This PR will fix the hacky way of adding the spacing between title and content. It will decrease the gap between title and content and also decrease the gap between the icon and the title.

## Changes

- Decrease the margin between the icon and the title
- Create a Column for the title and the content
- Use consistent gap between the title and the content

## Screenshots
Simulator Screen Shot - iPhone 11 Pro - 2020-05-06 at 21 55 13


lukesleeman (approved):
Ok, I can't comment on the flutter side of things, but I've run the PR and it looks good ✅

lib/view/disclaimer_view.dart
```diff
@@ -122,62 +122,65 @@ class _DisclaimerViewState extends State<DisclaimerView> {
...
                               ),
                             ),
-                            const Expanded(
-                              child: Text(
-                                'Please keep in mind\n',
-                                style: Styles.textH3,
-                              ),
-                            ),
+                            Expanded(
+                                child: Column(
```
workerbee22: Perhaps flutter format this with dartfmt ?

lib/view/disclaimer_view.dart
```diff
@@ -122,62 +122,65 @@ class _DisclaimerViewState extends State<DisclaimerView> {
...
+                                SizedBox(height: 10.0),
+                                Text(
+                                  'This is not a comprehensive source nor can we guarantee it is completely up to date at '
+                                  'the time of use 📱.\n\nIt is created using Western Health guidelines, informally '
+                                  'peer-reviewed and adapted, with permission, from College/Society guidelines.\n\n'
+                                  'WHAC19 gathers analytics and crash data which is used to improve the app.  For more information and to opt out, go to Information > Privacy\n\n',
+                                  style: Styles.textBody,
+                                )
+                              ],
+                            ))
```
workerbee22: Trailing comma here I think should also make sure dartfmt does it right so ),]

lib/view/disclaimer_view.dart
```diff
@@ -122,62 +122,65 @@ class _DisclaimerViewState extends State<DisclaimerView> {
...
                               ),
                             ),
-                            const Expanded(
-                              child: Text(
-                                'WHAC19 does not constitute official advice\n',
-                                style: Styles.textH3,
-                                softWrap: true,
-                              ),
-                            ),
+                            Expanded(
```
workerbee22, line 162: Ditto as per above for the hand emoji section.

workerbee22 (changes requested):
Just needs a litte formatting, otherwise good to go after that.

lib/view/disclaimer_view.dart
yimingc2: Yup. You are a great detective. Even though I format it manually, it still changes it back to this way.

lib/view/disclaimer_view.dart
yimingc2: Oh I know. It needs comma everywhere.

workerbee22 (approved):

PR #330, opened 2020-05-06
[Fix] Add spacing between items

### LINKED ISSUE #322: ICU cell bottom margins too large
This margin mismatch seems to only affect sections under ICU. ie, Ventilation, Daily Round, Proning... etc

The bottom margin is a lot larger than the top. 

Copy changes 011

yimingc2:
Resolved #322

## Abstract
This PR will fix the too large bottom margins for the cards in ICU cells (and the pages that use the same template).

## Changes

- Remove the bottom padding of each card
- Add spacing between cards

## Screenshots
Simulator Screen Shot - iPhone 11 Pro - 2020-05-06 at 22 43 46

## Things to note
Would be great if you know any other good ways to add spacing between elements. Using `Wrap` is bit hacky.

### DIFF
--- a/lib/widget/lists/icu_daily_round_steps_card.dart
+++ b/lib/widget/lists/icu_daily_round_steps_card.dart
@@ -11,54 +11,54 @@ class ICUDailyRoundStepsCard extends StatelessWidget {
 
   Widget _renderItems(ICUDailyRoundItem item) {
     return Row(
-        crossAxisAlignment: CrossAxisAlignment.start,
-        children: <Widget>[
-          Expanded(
-            flex: 10,
-            child: Padding(
-              padding: const EdgeInsets.fromLTRB(0, 0, 0, 0),
-              child: Column(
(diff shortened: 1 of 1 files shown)


lukesleeman (approved):
I've tested this and it looks good!

I also pushed a commit which fixes up the flutter format issue which was causing the build to fail.

Again - I can't speak to the flutter side of things, but its a ✅ from checking out and running the app

lib/widget/lists/icu_daily_round_steps_card.dart
```diff
@@ -87,7 +84,12 @@ class ICUDailyRoundStepsCard extends StatelessWidget {
         elevation: 0,
         child: <Widget>[
           _renderHeading(),
-          ...subsection.list.map((item) => _renderItems(item)).toList(),
+          Wrap(
```
workerbee22, line 87: Wrap is ok for now 👍It does the job.

workerbee22 (approved):
LGTM.

PR #332, opened 2020-05-07
Updates PDF icon to be more clear

marcedwards:
This is a very minor improvement to the PDF launcher button, to make it more clear. This is just an interim solution until we find a better solution for opening the PDFs. The updated image is the same size, so no code changes are required.

## Screenshot

Left = old. Right = new.

pdf-icon


lukesleeman (approved):
I've tested it out, and this looks good!  Its a bit more obvious than the old one

workerbee22 (approved):
LGTM simple asset replacement 👍

PR #333, opened 2020-05-07
Add place holder for last updated time

### LINKED ISSUE #288: Home and information screen changes
There are some modifications that need to be made to the home and information screens to better surface information:

- [x] Add new information section to home screen
- [x] Add phone icon to home screen
- [x] Update information screen, removing items
- [x] Add "last updated date" to the information screen
- [x] Add contributors under the Acknowledgements and references

# Add new information section to home screen
A new "information" section needs to be added to the home screen, after the ICU section.  It should contain two new buttons

**WH Definitive resources**
This needs to lead to a subscreen which has the first 3 buttons from the current information page.  These are:

1. WH Novel coronavirus website
2. WH anaesthetic microsite
3. WH ICU microsite

**Additional resources**
This is the screen which the existing "Additional resource" button in the existing information screen leads to

# Add phone icon to home screen
On the home screen, we need to add a new phone icon to the right of the current "information" icon.  The phone icon should take you to the website which the existing "Contact numbers" button in the existing information screen leads to ( http://wh.cyphix.net/ )

# Update information screen
The first 5 buttons in the information screen can now be removed.  The only items remaining on the information screen should be:

- App feedback
- Disclaimer and conditions of use
- Acknowledgements and references
- Third party licenses
- WHAC19 Icon to remain present

# Add "last updated date" to the information screen
So that users can gauge how up to date the information in the app is we need to show the date their build was created.  Hopefully this should be prominently displayed close to the app icon, in the information screen.  Given the info screen has been re-ordered it might even be nice to give it a little more ✨  by moving the icon around or otherwise shuffling around things.

From a coding point of view, ideally this date should be based off the build time, so we don't have to update a string every time we do a release.

# Add contributors under the Acknowledgements and references
We want to add a new section to the Acknowledgements and references screen which lists all the contributors ❤️

yimingc2:
Resolved #288

## Abstract
This PR will add a placeholder for the last updated time in env.dart. Then CI should update this variable in a format of `10:17pm 07 May 2020`

## Changes

- Add last updated date placeholder
- Display last updated date if there's any in the info view

## Screenshots
Screenshot_1588854003

## Things to note
If CI has not added the last updated time, then CI might overwrite `env.dart` and variable `lastUpdatedTime` will disappear. And it could result in build failure. @maks

### DIFF
--- a/lib/view/info_view.dart
+++ b/lib/view/info_view.dart
@@ -83,7 +83,6 @@ class InfoView extends StatelessWidget {
   static Widget _buildSpacer({double height = 16}) => Container(height: height);
 
   Widget _buildLastUpdatedDate() {
-
     String formatTimestamp(int timestamp) {
       final format = DateFormat('h:mma dd MMMM yyyy');
       final date = DateTime.fromMillisecondsSinceEpoch(timestamp * 1000);
--- a/lib/view/info_view.dart
+++ b/lib/view/info_view.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:intl/intl.dart';
 import 'package:package_info/package_info.dart';
 
 import '../env.dart';
@@ -82,13 +83,22 @@ class InfoView extends StatelessWidget {
   static Widget _buildSpacer({double height = 16}) => Container(height: height);
 
   Widget _buildLastUpdatedDate() {
-    if (lastUpdatedTime.isEmpty) {
+
+    String formatTimestamp(int timestamp) {
+      final format = DateFormat('h:mma dd MMMM yyyy');
+      final date = DateTime.fromMillisecondsSinceEpoch(timestamp * 1000);
+      return format.format(date);
+    }
+
+    final timestamp = int.tryParse(lastUpdatedTime);
+
+    if (timestamp == null) {
       return Container();
     } else {
       return Container(
         margin: const EdgeInsets.all(8.0),
         child: Text(
-          'Last updated $lastUpdatedTime',
+          'Last updated ${formatTimestamp(timestamp)}',
           textAlign: TextAlign.center,
           style: Styles.textFooter,
         ),
--- a/lib/view/info_view.dart
+++ b/lib/view/info_view.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:package_info/package_info.dart';
 
+import '../env.dart';
 import '../hard_data.dart';
 import '../routes.dart';
 import '../strings.dart';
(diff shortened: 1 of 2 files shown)


workerbee22 (approved):
LGTM

lib/view/info_view.dart
```diff
@@ -80,6 +81,21 @@ class InfoView extends StatelessWidget {
...
   static Widget _buildSpacer({double height = 16}) => Container(height: height);
 
+  Widget _buildLastUpdatedDate() {
+    if (lastUpdatedTime.isEmpty) {
+      return Container();
+    } else {
+      return Container(
+        margin: const EdgeInsets.all(8.0),
+        child: Text(
+          'Last updated $lastUpdatedTime',
```
maks: Probably want the timestamp format to include the timezone? 
The default unix date format is eg `Fri 08 May 2020 14:49:33 AEST` can we just use that?

yimingc2: If it takes a lot effort to do time formatting then I think this is okay for now. In the design, it is "9:41am 26 April 2020". We can do as close as possible. Can it convert to the local time without timezone?

lukesleeman:
So, I pulled down the last build from the #covid-19-app-builds channel, but unfortunately I can't see any last update date in it.  Not sure if its supposed to be showing up yet?

yimingc2:
> So, I pulled down the last build from the #covid-19-app-builds channel, but unfortunately I can't see any last update date in it. Not sure if its supposed to be showing up yet?

Yeah, that's okay. I just add a placeholder for the last updated time. CI will assign a value(the updated time) for this placeholder when Codemagci is running and after CI build successfully, we will be able to see the last updated date. 

Maks is working on the CI part atm.

workerbee22:
its a one liner in the pre script in codemagic. After thats in it should
work fine.

maks:
Ok I've added to CI the timestamp as seconds after epoch, @yimingc2  you can then just easily format it as needed on Dart side.  I've also added this to PR build workflow and kicked off a new PR build with this PR branch to test it out.

yimingc2:
Hi @maks @lukesleeman , I've tested out on Android side with the CI build. It works well. Thanks for @workerbee22 advice on timestamp.
Screenshot_1588992441

lukesleeman (approved):
I've downloaded the CI build and confirmed that it contains the build date!  I've also run this locally based off source and can confirm that everything is fine in a dev enviorment when the value isn't set.

Can't speak to the flutter side of things, but based on my testing this PR gets the ✅ 

Good job @yimingc2 and @maks  🙌

yimingc2:
Nice. Thanks Luke. I’ll test it out again after this version is uploaded to
TestFlight. Just in case.

PR #336, opened 2020-05-11
Fix infographics SVG transparency

marcedwards:
The “PDF” portion of this icon was white, where it should have been transparent.

## Screenshots

before-after


lukesleeman (approved):
I've tested out the changes and this looks good

workerbee22 (approved):
LGTM ... just an asset change, with same name 👍

PR #339, opened 2020-05-15
Onboarding Screens

### LINKED ISSUE #113: Onboarding screens
We have the disclaimer which pops up when the app launches - but of course nobody really reads that.  There are a few key things it might be good to make really obvious to users when they first launch:

☑️ What is actually in the app: Info for aesthetic and critical care doctors and nurses.  PPE, intubation, etc
🙅‍♀️The app is informational.  Don't rely on it exclusively, confirm things against actual guidelines, use your judgment

Mike suggested some sort of onboarding pager thing they swipe through.

This may also be a good way to solve issue #108  - support for other hospitals.  On the last page we could ask them what hospital they work in.  For the time being it could be WH only, but the idea is we could add more to the list.

workerbee22:
Implement onboarding screens at app startup as per [issue 113](https://github.com/Western-Health-Covid19-Collaboration/wh_covid19_app/issues/113).

Requires:
- [x] Modified Disclaimer content view to move some content to onboarding screens
- [x] New iconography assets for onboarding screens and disclaimer
- [x] Onboarding to run only on first run of the app and save that state to storage
- [x] Onboarding work with Disclaimer view which may be updated independently 
- [x] Add onboarding screens for key message and content

## Changes

- Disclaimer view update 
- New onboarding iconography assets
- Add onboarding screens and persistent storage values
- Change to init router / routing app behaviour
- Content in strings, constants and hard_data files etc.
- Tests updated

## Screenshots

Updated Disclaimer screen on app startup:

image

Onboarding screens at ap startup:

image

## Things to note

- To test this feature, uninstall then re-install app to your device to force Disclaimer and then Onboarding screens to be presented.
- Disclaimer content updated by moving some to onboarding screens and current version is 7
- Disclaimer screen at app startup style matches the onboarding screens at startup
- Possible first screen combinations are:

1. Users first run after app install, screen shown are: Disclaimer, Onboarding and then Home screen.
2. User has app installed then upgraded the app AND Disclaimer has changed in the new version, screens shown are: new Disclaimer screen only (onboarding screens NOT shown again).
3. User is using the app normally, screens shown are: Home screen only.

- Disclaimer and Onboarding screen urls are present for reference only and NOT enabled as links, so as not to encourage users to go to those pages whilst onboarding into the app.


greggmiller (approved):
LGTM

workerbee22:
No one around, but @greggmiller has reviewed this change and approved. I've also tested on many devices and configurations, so going to merge so we can produce Beta 4.

PR #350, opened 2020-05-28
[Update] Update iOS & Tooling Support

workerbee22:
General update of the iOS side of the project now that we are back on Flutter stable channel:

- Xcode 11.5 and iOS 13.5 re-test
- Minor packages updates to fix minor issues on iOS
- Tweak bottom of screen spacing for Disclaimer and Onboarding for iOS 13

## Changes

- Updated to support several package updates to remove minor issues on iOS.
- Updated and retested to support Xcode 11.5 and check iOS 13.5 / 12.4 support.
- Retested on Android to ensure changes ok
- Adjusted spacing on Disclaimer and Onboarding screens at the bottom to avoid being too close to the iOS 13 controls.
- Updated Contributing instructions to reflect current versions.

## Screenshots

Slightly more vertical space to avoid being too close to iOS 13 control. Still looks fine on Android

image

image

## Things to note

- codemagic workflows changed to reflect Xcode 11.5 and retested
- Note kept cocoapods at 1.9.1 as newer version seems to add no fixes or features.
- Release notes not updated as no new features or functional changes

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -12,20 +12,20 @@ dependencies:
     sdk: flutter
   package_info: ^0.4.0+18
   chewie: ^0.9.10
-  video_player: ^0.10.11
-  flutter_widget_from_html: ^0.3.3+3
+  video_player: ^0.10.11+1
+  flutter_widget_from_html: ^0.3.3+4
   matrix_gesture_detector: ^0.1.0
-  url_launcher: ^5.4.7
+  url_launcher: ^5.4.10
   flutter_svg: ^0.17.4
-  shared_preferences: ^0.5.7+2
+  shared_preferences: ^0.5.7+3
   intl: ^0.16.1
   photo_view: ^0.9.2
   provider: ^4.1.2
-  firebase_core: ^0.4.4+3
-  firebase_analytics: ^5.0.11
+  firebase_core: ^0.4.5
+  firebase_analytics: ^5.0.14
   sentry: ^3.0.1
   device_info: ^0.4.2+4
-  native_pdf_view: ^3.4.0
+  native_pdf_view: ^3.6.0
 
 dev_dependencies:
   flutter_test:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ You will need to install and setup Flutter to develop this application. You can 
 - the installation of dependencies
 - running the application
 
-### **STABLE CHANNEL USAGE**
+### Developer Setup
 
 We are currently using the Flutter stable channel (currently v1.17.1) for app development, so once you have the Flutter
 SDK set-up make sure you are on the stable channel:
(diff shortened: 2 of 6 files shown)


pubspec.yaml
```diff
@@ -12,20 +12,20 @@ dependencies:
     sdk: flutter
   package_info: ^0.4.0+18
   chewie: ^0.9.10
-  video_player: ^0.10.11
-  flutter_widget_from_html: ^0.3.3+3
+  video_player: ^0.10.11+1
+  flutter_widget_from_html: ^0.3.3+4
   matrix_gesture_detector: ^0.1.0
-  url_launcher: ^5.4.7
+  url_launcher: ^5.4.10
```
yimingc2, line 18: Do we need to manually change the minor version? ^5.4.7 is equivalent to '>=5.4.7 <6.0.0' (next non-broken version). I think if we want to upgrade the dependency version, we can just do `flutter pub upgrade`

pubspec.yaml
```diff
@@ -12,20 +12,20 @@ dependencies:
...
   flutter_svg: ^0.17.4
-  shared_preferences: ^0.5.7+2
+  shared_preferences: ^0.5.7+3
   intl: ^0.16.1
   photo_view: ^0.9.2
   provider: ^4.1.2
-  firebase_core: ^0.4.4+3
-  firebase_analytics: ^5.0.11
+  firebase_core: ^0.4.5
```
yimingc2, line 24: If the major version number is 0, then it means the version is not stable yet. ^0.4.4 is equivalent to '>=0.4.4 <0.5.0'.

yimingc2 (approved):
It's great to update to the latest version! It's pretty good. Just little dependency version range thing that can take a look.
ps. Spent a lot of time to update every thing.

pubspec.yaml
workerbee22: No we don't as the `^` does that for us. I just do this as a manual check ie. I check the newer releases and see what has changed, so I can figure out what I might need to re-test.

So the manual update of minor version is just so we know we've looked at the changes up to this minor version and they seem ok for this project. Otherwise yes we can always do a `flutter pub upgrade` and so the versions in use are actually the ones in `pubspec.lock`

pubspec.yaml
workerbee22: yep see above .

workerbee22:
So yep the manual update of minor version isn't actually required, but it is just so we know we've looked at the changes up to this minor version and they seem ok for this project. I usually don't do this until we are close to a release as it helps figure out what changes are in recent package releases and what might need testing prior to the release. Rather than just blindly getting the later minor version updates .... which I prefer not to do when we are about to release to users.

Otherwise yes we can always do a `flutter pub upgrade` and so the versions in use are actually the ones in `pubspec.lock`.

@yimingc2 as discussed, given you are probably one of the only other Devs who checks the iOS version, I'll take your approval as OK to go 👍

